PR #10, opened 2020-08-26
adding transit provider list

amymapsmith:


thekaveman:
Looks great @amymapsmith! Do you want to shorten the file name to keep it simple? `agencies.csv` would do.

thekaveman (changes requested):
Hi Amy, this is getting close! Can you please make the following small changes/cleanups?

## Remove old files

These files aren't going to be used on the website and so can be deleted: 

* `data/Map.svg`
*  `data/agency_list_transit_store.csv` (duplicate of `data/agency.csv`)

## Cleanup `data/map.geojson`

This file could be slimmed down to improve load time:

* Remove `Feature`s with `geometry.coordinates == []` (there are at least 2 of these).
* Remove `properties` not needed for the map. CC @carvin working on the visuals and map gradient. From what I can tell, the only `properties` used right now are:
  * `county`
  * `num_providers`
* Try to remove formatting like unnecessary spaces and/or line breaks. Every byte counts!

thekaveman (approved):

PR #53, opened 2020-09-25
Update counties.geojson

amymapsmith:
Fixing Imperial County count with addition of Yuma County Area Transit


thekaveman:
Sorry, I didn't realize you had updated the Yuma entry in the providers table in the other PR! Made a bit of a mess of this one, tried to clean it up...

PR #58, opened 2020-09-28
Update dictionary.csv

amymapsmith:
Additional data notes for service counties field.


thekaveman (changes requested):
Hi @amymapsmith - it looks like the notes contain a comma, so the new value needs to be wrapped in quotes or else it throws off the CSV as a whole.

thekaveman (approved):
👍

PR #80, opened 2020-10-01
Updated provider list

amymapsmith:
Updated provider list


thekaveman (approved):
Thanks @amymapsmith. Next time you can submit the updates to providers and counties in the same PR; this way our preview will have the unified data (vs. 2 previews each with half the updated data).

PR #81, opened 2020-10-01
Add files via upload

amymapsmith:
Provider total by county to join to Geojson


thekaveman:
We decided not to use this for now since it will require rewriting the map.

PR #82, opened 2020-10-02
Delete counties.geojson

amymapsmith:
To be replaced by updated geojson.


thekaveman:
We don't need this PR since the file was updated in #83, and we will use that.

PR #83, opened 2020-10-02
Update counties.geojson

amymapsmith:
New geojson with updated num_providers.


thekaveman (approved):
I left the same note on #80 - next time you can update these files in the same PR, so we have unified data for the preview.

Issue #269, opened 2021-05-10
Hyperlink vendor names on MobiMart Contact page.


harrisonpeck:
The Elavon EPAY contract is now fully amended, signed, and live. We need to link to this on MobiMart. And since we're adding a link for Elavon, Gillian would also like to link to FiServ as well. So:

On the "Contact us" page, where it says "Elavon", please link to this URL: https://caleprocure.ca.gov/PSRelay/ZZ_PO.ZZ_CTR_SUP_CMP.GBL?Page=ZZ_CTR_SUP_PG&Action=U&SETID=STATE&CNTRCT_ID=5-10-99-02

Where it says "Fiserv", please link to this URL: https://caleprocure.ca.gov/PSRelay/ZZ_PO.ZZ_CTR_SUP_CMP.GBL?Page=ZZ_CTR_SUP_PG&Action=U&SETID=STATE&CNTRCT_ID=5-10-99-01

![image](https://user-images.githubusercontent.com/80477201/117688474-cabc3c80-b186-11eb-9ab6-564904fd99f1.png)

Underneath that line, insert "Questions? Contact Cal-ITP using the form below."

Under "Fare calculation software" where it says "Current vendors--coming soon. Sign up below for updates", please change this text to read, "View the draft RFP", and hyperlink the words "draft RFP" to this URL: https://caleprocure.ca.gov/event/77601/0000019394. 

Then please do the same under "Fare validators" as well.

allejo:
Should these links open in new tabs or in the current window?

allejo:
Oops, closed by #270

PR #270, opened 2021-05-10
Add links for vendors on Contact page

allejo:
See #269


thekaveman (changes requested):
I saw your question on the issue; I think we should have these open in a new tab. Especially because that Cal E-Procure website is a back-button black hole.

PR #275, opened 2021-05-26
Reintroduce Sass and significantly simplify styles

allejo:
This is a major overhaul of all of the styles used on the website and reintroduces Sass. This overhaul does **not** result in a pixel perfect equivalent to what's currently in master but it's as close as I can get it with simplified styles.

This is a huge PR and I have split things up into appropriate commits, so I'd recommend reviewing the changes one commit at a time.

## Todo

- [x] Provider Map
- [x] Fix potentially broken navigation bar


thekaveman (changes requested):
Just an initial comment to note some minor discrepancies (without yet going through the file changes individually):

* (Desktop) main nav is squished to the right, should stretch across the width of page
* (Desktop) Contact page/form is too wide, same with Contracts
* Need some padding-bottom on H1's on content pages: About, Contact, Glossary

thekaveman (changes requested):
Since you're overhauling, may as well normalize content pages H1 center-aligned with border-bottom

* Contracts
* Provider map
* About
* Contact
* Glossary

thekaveman (changes requested):
This is looking really great! Couple more fixes:

* Thanks for normalizing links, but we lost some readability; please add back the underline
* Provider map: too much space between header nav and title
* About, Contact: need a little more space between content and footer nav

thekaveman (changes requested):
Almost... the link on the homepage under *Payment system RFP, hot off the press!* is not styled the same as the others. 

Is there another way to do it besides styling links under the `.text-scope` class? Seems easy to forget.

allejo:
As we discussed, switched over to a opt-out approach for anchor links in ff4bed225e3d4dfe018bd42cf7c444cc718f3e02. We're now targeting `.main-content a` and styling them the same (with the exception of anchors styled as buttons). Doesn't look like we have any components that require special styling of anchors, so we're good with an opt-out approach.

_sass/overrides/_main-content.scss
```diff
@@ -0,0 +1,5 @@
+.main-content a {
+  &:not(.btn) {
+    font-weight: bold;
```
thekaveman, line 3: The CA State Template adds this ridiculous link underline with a background image, and we need to wipe that out. Here's what we had before, need something similar/equivalent:

```css
background-image: none !important;
background-repeat: no-repeat !important;
background-size: unset !important;
background-position: unset !important;
background-origin: unset !important;
border-bottom: solid 2px $primary;
```

_sass/pages/_provider-map.scss
```diff
@@ -0,0 +1,52 @@
+$map-height-sm: 460px;
+$map-height-md: 560px;
+
```
thekaveman, line 3: Can we get a little extra space below the intro content please? Something like (this is probably a horrible rule, but you get the idea):

```sass
.provider-map .main-primary > .container:first-of-type {
    padding-bottom: 1rem
}
```

_sass/pages/_provider-map.scss
```diff
@@ -0,0 +1,52 @@
...
+.provider-data-download-button {
+  width: 230px;
+}
+
+.provider-california-map {
```
thekaveman, line 8: Speaking of the annoying link underline from CA State Template...

There is a hidden link control coming from Mapbox, but we still get the underline effect:

image

I think something like this would solve it:

```css
.provider-california-map {
    .mapboxgl-ctrl-bottom-left {
        display: none;
    }
}
```

thekaveman (changes requested):

thekaveman (approved):
🙌 fantastic work @allejo!

PR #281, opened 2021-06-06
Track downloads of Provider Data

### LINKED ISSUE #273: Understand data package downloads
We have Google Analytics for basic metrics like pageviews and time spent on page. We'd also like to know how many downloads of the [provider data package](https://github.com/cal-itp/transit-stop/tree/master/metadata/providers) there are.

This could be done with a [GA Event](https://developers.google.com/analytics/devguides/collection/analyticsjs/events) on click of the [download button](https://github.com/cal-itp/transit-stop/blob/master/provider-map.html#L81) on our Providers page.

allejo:
Technically, this unique enough and small enough that it can just be added to the Default layout. Though, I'm not sure if we'd be finding more things to track and just do things on a per-page basis. At the moment, I'm abusing front matter to place all of our related Google function calls together in a single script tag.

Fixes #273

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -62,6 +62,8 @@
     gtag('js', new Date());
 
     gtag('config', '{{site.google_analytics_id}}');
+
+    {{ page.gtags_script }}
   </script>
 </body>
 </html>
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
Yeah this is fine for now

PR #282, opened 2021-06-08
fix(analytics): use gtag script instead of ga

thekaveman:
Events were not being registered in Google Analytics, need to use the Universal Analytics syntax:

https://developers.google.com/analytics/devguides/collection/gtagjs/events


thekaveman:
@allejo FYI, I had to make one small change to #281 after checking in GA and not seeing any events being registered.

PR #320, opened 2022-02-15
Add new resources page, route, links

### LINKED ISSUE #311: Mock up a Resources page
Let's create a preview (via the standard PR process) for what the Resources page will look like.

We'll just go with a fairly bare-bones page for now (no illustrations, no interactive features - just content):

- [x] path should be `/resources`
- [x] title should be `Resources`
- [x] using the standard navbar / footer as other pages
- [x] one paragraph of intro text (lorem ipsum for now)
- [x] sub-sections for e.g. "Getting Started" (lorem ipsum placeholders for 3-4 other sub-section headings for now)
- [x] one or two sentences of explainer text under each sub-section (lorem ipsum for now)
- [x] one or more links under each sub-section (`href="#"` until we have the real files) - maybe a list? or table?

machikoyasuda:
closes #311 

## What this PR does
- Create a Resources page
- ~Add Resources link to the Header~
- Resource page is created with an HTML file ~, with data from `resources.yml`~
- ~Write code that iterates over `resources.yml` to show the resource name, resource explainer, then the links.~

## Screenshots
image
image


resources.html
```diff
@@ -0,0 +1,23 @@
...
+  </p>
+</section>
+
+<section>
+  {% for resource in site.data.resources %}
+    <h2>{{ resource.name }}</h2>
+      <p>{{resource.explainer}}</p>
+      <ul>
+        {% for link in resource.links %}
+          <li><a href={{ link }}>Link</a></li>
```
thekaveman, line 19: Here let's use the link `href` as the text, rather than repeating `Link` for each one.

Ideally we'll have files that we can parse into meaningful names e.g. `/getting-started/contactless-payments.pdf` in the bucket could be rendered under `Getting Started` with the link text `Contactless Payments`. We can worry about that when we move into automating the GCS pull.

thekaveman (changes requested):

resources.html
machikoyasuda: Now resolved https://github.com/cal-itp/mobility-marketplace/pull/320/commits/fc75434cf8a375d826d0343ab19df37833932562

image

angela-tran (approved):

thekaveman (approved):
OK this looks good for the starting mock! @machikoyasuda can you please post a note in #websites with the preview link and tag Erika to review this page?

⚠️We don't want to merge this until we have the actual data here! ⚠️

thekaveman:
Updated based on some feedback from Slack. Since this seems to be more fluid right now, I also removed the data file and went back to hardcoded HTML. Once we have a more stable structure here we can revert to the data file.

angela-tran:
The approved text content has been added, but we are still waiting for the docs to be approved. Then we can upload them and link to them.

⚠️We don't want to merge this until we have the actual **links** here! ⚠️

PR #325, opened 2022-02-24
Hide footnotes

### LINKED ISSUE #322: Contracts Table: shows "1 Available to CA agencies only" footnote on all contracts; pls. remove from 2 new Product Categories
When applying a Product Category "sort" to [Contracts Table](https://www.camobilitymarketplace.org/contracts/view), Gillian would like the "1. Available to California agencies only." _footnote_ **removed** from these two Product Category sorts:
• [**Transit Processor**](https://www.camobilitymarketplace.org/contracts/view?contracts-filter-product=Transit%20Processor)
• [**Payment Acceptance Devices**](https://www.camobilitymarketplace.org/contracts/view?contracts-filter-product=Payment%20Acceptance%20Devices)

Images from Gillian to illustrate what we need to remove.
State of California Contracts  California Mobility Marketplace
State of California Contracts  California Mobility Marketplace

**PLEASE NOTE:** Footnote is accurate when it appears under the _entire table_ of contracts or when it appears under the _[Payment Processor](https://www.camobilitymarketplace.org/contracts/view?contracts-filter-product=Payment%20Processor)_ Product Category sort.

Thanks!

angela-tran:
Closes #322

image

image

image

### DIFF
--- a/contracts/view.html
+++ b/contracts/view.html
@@ -7,6 +7,7 @@ class_name: no-banner
 
 <script defer src="https://unpkg.com/tabulator-tables@4.9.3/dist/js/tabulator.min.js"></script>
 <script defer src="/scripts/smart-table.js"></script>
+<script defer src="/scripts/contracts/footer.js"></script>
 
 <div class="container container--skinny">
   <article class="mb-5">
@@ -92,7 +93,23 @@ class_name: no-banner
     <footer>
       <ol class="footnote-listing">
         {% for footnote in site.data.contracts.footnotes %}
-          <li id="fn-{{ footnote[0] }}">
+          {% assign products = "" | split: "" %}
+          {% assign footnote_id_str = footnote[0] | append: "" %}
+          
+          {% comment%}
+          Build up product data-attribute string for each footnote 
+          {% endcomment %}
+          {% for entry in site.data.contracts.entries %}
+            {% assign entry_footnotes = entry.footnotes | join: "," | split: "," %}
+            {% if entry_footnotes contains footnote_id_str %}
+              {% unless products contains entry.product %}
+                {% assign p = entry.product | join: "," | split: "," %}
+                {% assign products = products | concat: p %}
+              {% endunless %}
+            {% endif %}
+          {% endfor %}
+
+          <li id="fn-{{ footnote[0] }}" data-products="{{ products | join: "," }}">
             {{ footnote[1] }}
           </li>
         {% endfor %}
--- a/contracts/view.html
+++ b/contracts/view.html
@@ -110,7 +110,8 @@ class_name: no-banner
           {% endfor %}
 
           <li id="fn-{{ footnote[0] }}" data-products="{{ products | join: "," }}">
-            {{ footnote[1] }}
+            <span class="footnote-number">{{ footnote[0] }}.</span>
+            <span>{{ footnote[1] }}</span>
           </li>
         {% endfor %}
       </ol>
--- a/scripts/contracts/footer.js
+++ b/scripts/contracts/footer.js
@@ -6,7 +6,7 @@
 const updateFooter = filters => {
   const productFilters = filters.filter(f => f.field === 'product' && f.type === 'like');
   
-  const footnotes = $('footer').children('ol').children('li');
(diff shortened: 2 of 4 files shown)


contracts/view.html
```diff
@@ -92,8 +93,25 @@ <h1 class="page-header">State of California Contracts</h1>
...
+          {% assign products = "" | split: "" %}
+          {% assign footnote_id_str = footnote[0] | append: "" %}
+          
+          {% comment%}
+          Build up product data-attribute string for each footnote 
+          {% endcomment %}
+          {% for entry in site.data.contracts.entries %}
+            {% assign entry_footnotes = entry.footnotes | join: "," | split: "," %}
+            {% if entry_footnotes contains footnote_id_str %}
+              {% unless products contains entry.product %}
```
thekaveman, line 105: So this `{% unless %}` guards against adding the same product twice, when multiple product entries have the same footnote?

scripts/contracts/footer.js
```diff
@@ -0,0 +1,29 @@
...
+const updateFooter = filters => {
+  const productFilters = filters.filter(f => f.field === 'product' && f.type === 'like');
+  
+  const footnotes = $('footer').children('ol').children('li');
+  footnotes.each((i, elem) => {
+    const _dataAttr = $(elem).attr("data-products");
+    
+    const footnoteProducts = _dataAttr ? _dataAttr.split(",") : [];
+    const selectedProducts = new Set(productFilters.map(f => f.value));
+    const intersection = new Set(footnoteProducts.filter(p => selectedProducts.has(p)));
```
thekaveman, line 15: Very unfortunate there is no built in `Set` intersection operator!

scripts/smart-table.js
```diff
@@ -92,6 +96,8 @@ $(function () {
           pillContainer.prepend($pill[0]);
           pillFilters.add(cellValue);
         }
+
+        smartTableCallbacks.dataFiltered.forEach(cb => cb(filters));
```
thekaveman, line 100: Using a list of callbacks here is really clever! Gives us a nice way to expand this in the future if the need arises.

scripts/contracts/footer.js
```diff
@@ -0,0 +1,29 @@
+/*
+  `updateFooter` is specific to the Contracts table in `contracts/view.html`.
+  
+  It integrates with `smart-table.js` to be called when data is filtered.
+*/
+const updateFooter = filters => {
+  const productFilters = filters.filter(f => f.field === 'product' && f.type === 'like');
+  
+  const footnotes = $('footer').children('ol').children('li');
```
thekaveman, line 9: I think my only requested change would be this line - the `$('footer')` selector seems too broad (we have 2 `<footer>` elements on the page already).

Can we use the footnotes list selector instead? `ol.footnote-listing`. That way we're sure to get the list of footnote items.

thekaveman (changes requested):
Just one minor change requested. 

Otherwise this is super clever @angela-tran! I love the implementation 🚀

scripts/contracts/footer.js
angela-tran: Good suggestion 😄 - I changed it to `$('ol.footnote-listing').children('li')`

contracts/view.html
angela-tran: Yeah, unfortunate that they don't have a way of doing `not contains`.

scripts/contracts/footer.js
angela-tran: https://github.com/cal-itp/mobility-marketplace/pull/325/files#diff-4b3572daf35019e63906407ff17148ff60f18b7cb4b0825555c1131e3ed40b5fR9

thekaveman (approved):
Nitpick: We should have a consistent quote character in JS files, we're mixing single and double quotes in the `scripts/contracts/footer.js` file.

(My preference is always double)

angela-tran:
> Nitpick: We should have a consistent quote character in JS files, we're mixing single and double quotes in the `scripts/contracts/footer.js` file.
> 
> (My preference is always double)

Thanks for pointing this out - I will go ahead and make it consistent here!

PR #331, opened 2022-03-04
Track resource downloads

### LINKED ISSUE #330: Track resource downloads in GA
On the [Provider Map page](https://www.camobilitymarketplace.org/provider-map), we send a `download` event when a user clicks the button to download the data package.

The code to do this is pretty simple and is currently defined in the front-matter in [`provider-map.html`](https://github.com/cal-itp/mobility-marketplace/blob/main/provider-map.html#L5):

```js
function handleDownloadClick() {
    gtag('event', 'download', {
      event_category: 'downloads',
      event_label: 'Provider Data Package',
    });
  }
```

Let's make this function more generic, to accept a custom `event_label` and make it available to all pages via the [default layout](https://github.com/cal-itp/mobility-marketplace/blob/main/_layouts/default.html#L70).

Then in `resources.html`, on each link for the download items, add this function call to the `onclick` attribute.

This also ties in with #328 and will give us further insight into how much traffic these files are getting.

angela-tran:
Closes #330

Question for @thekaveman : is there a way for me to test that events are being sent to Google Analytics?

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -67,7 +67,12 @@
 
     gtag('config', '{{site.google_analytics_id}}');
 
-    {{ page.gtags_script }}
+    function handleDownloadClick(event_label) {
+      gtag('event', 'download', {
+        event_category: 'downloads',
+        event_label: event_label,
+      });
+    }
   </script>
 </body>
 </html>
--- a/provider-map.html
+++ b/provider-map.html
@@ -2,13 +2,6 @@
 layout: default
 title: Provider Map
 class_name: provider-map no-banner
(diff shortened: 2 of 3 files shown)


thekaveman:
@angela-tran Yes! I was able to test the preview site in Google Analytics by using the `CA Mobility Marketplace > Unfiltered Data` View, which _does not_ filter by host name (like `Production`):

image

Then I looked under `Realtime > Events` and when I clicked the button to download the data package, an event was recorded!

image

thekaveman (approved):

PR #379, opened 2022-10-25
docs(README): add server address to instructions

machikoyasuda:
I was unable to get the app open on my machine today. I added this one line of instruction to be more explicit what server address to open in the browser. 

Please check to see if these instructions work for you on your machines.


thekaveman (approved):
Thanks!

PR #434, opened 2023-03-27
Fixes: Contact Form

### LINKED ISSUE #433: Fixes: Contact Form
These are edits from @segacy1: 

- I’ve added updated designs/states for input fields, checkmarks, and radio buttons [here](https://www.figma.com/file/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=4268-17731&t=1RYcAOXwQRqdeOgw-4).
- Can we update the padding to be the following (reflected in figma and I can add pics or tag you in specific spots if that’s clearer):
- Radio button selector q's:
- 8px between checkmark/radio button and text
- 4px between each selection
- 8 px between question text and input fields
- 24px btwn each question (dsk), MBL: 20px btwn each question (mbl)
- Can we update the behavior for missing field inputs? Ideally, an alert won’t appear unless the user tries to submit without completing all required forms. If they do, the missing fields will display in their error state and (if possible) the page will jump back up to the first missed field.
- Yes please let’s match the nav/content sitch here to the how-to page!
- Please swap the hero image with this one and hide hero image on mobile:

machikoyasuda:
closes #433 

### Style
- Update `primary-black` to `#323a45`
- Create `.fs-xs-5` util class to represent "fs" (font size) of 5 (that's 20px) only on xs size, and then fs-4 on larger screens
### Hero container
- Update image file
- For desktop: Hero container starts at the same point as the How To pages. Currently, the Hero container starts too low.
- For desktop: Text column and Image column now have different alignment/widths, confirmed with Segacy.
- For mobile: Hide image

### Form
- Hide the missing field input behavior entirely
- Radio button: The check style is a full circle
- Form on mobile: Text size goes down to 20px for radios, checkboxes and inputs. 
- Form on mobile: Checkboxes and radio buttons should be vertically aligned with label. Adjust padding.

## What this PR does not do
In order of priority:
- Confirmation page: Hide the input #435 
- Style validation message #436 
- Checkbox button: Code out custom checkbox button #437


machikoyasuda:
Finally ready for review @segacy1 @angela-tran @thekaveman

segacy1:
This looks great @machikoyasuda ! One small edit—could you match the spacing between the radio buttons for _Does your agency currently produce GTFS data_ to what it is for the _I am a_ radio buttons?

Aside from that, it's good to go!

machikoyasuda:
@segacy1 has looked over and approved the design, so this PR is also ready for final code review @angela-tran @thekaveman

src/_sass/_utilities.scss
```diff
@@ -17,18 +17,35 @@ $mm-dark-tan: rgb(246, 243, 237);
...
   color: $gray-4;
 }
 
+.fs-xs-5 {
+  font-size: calc(20rem / 16);
+
+  @include media-breakpoint-down(md) {
+    font-size: calc(24rem / 16);
+  }
+}
```
thekaveman, line 40: Would it make more sense to have the rules for this class wrapped in a media query specifically for the `xs` size? 

Then we don't have to undo/override anything else in this class, and it won't interact with any other sizing rules for larger screens elsewhere.

In other words, this class "says" it is for `xs-5`, but then it also has rules for `md-4`, which seems unnecessary?

machikoyasuda:
Rebase main

src/_sass/_utilities.scss
machikoyasuda: Resolved here https://github.com/cal-itp/mobility-marketplace/pull/434/commits/b191d9a929c41e74ca75f1be0467535880c9aa40

machikoyasuda:
Ready for re-review @thekaveman

thekaveman (approved):
Looks great!

PR #489, opened 2023-11-14
Refactor: update existing sections

angela-tran:
Closes #467 

### Top section
- [x] Update "Explore this site to find support..." copy
- [x] Remove copy "Contracts update!"
- [x] Remove copy "U.S. public transit providers, we have news for you..."
- [x] Add View Contracts button in new style

### Lower sections
- [x] Remove map section
- [x] Remove "Modernize your fare collection system" section
- [x] Implement design for "What is Cal-ITP" section
- [x] Implement design for "Get in touch" section

PR #491, opened 2023-11-15
Feat: Modernize Fleet section

### LINKED ISSUE #466: Homepage: Create and add the Modernize your fleet section with 4 agency logos.
https://www.figma.com/file/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?type=design&node-id=6819-1391&mode=design&t=nr3BFzLE4p6yUjPx-0

image

angela-tran:
Closes #466 

Based off changes from #489

### DIFF
--- a/src/_sass/pages/_go-contactless.scss
+++ b/src/_sass/pages/_go-contactless.scss
@@ -38,7 +38,7 @@ $gc-blue: rgb(235, 248, 252);
 }
 
 .go-contactless-benefit-card,
-.go-contactless-agency {
+.agency {
   border-radius: 6px;
   box-shadow: 0 4px 4px rgba(0, 0, 0, 0.1);
   overflow: hidden;
@@ -49,7 +49,7 @@ $gc-blue: rgb(235, 248, 252);
   padding: 1rem;
 }
 
-.go-contactless-agency-container {
+.agency-container {
   display: grid;
   grid-gap: 2rem;
   padding: calc(32rem / 16) 0;
@@ -63,7 +63,7 @@ $gc-blue: rgb(235, 248, 252);
   }
 }
 
-.go-contactless-agency {
+.agency {
   display: flex;
   justify-content: center;
   align-items: center;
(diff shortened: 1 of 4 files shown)


src/_sass/pages/_go-contactless.scss
```diff
@@ -38,14 +38,14 @@ $gc-blue: rgb(235, 248, 252);
 }
 
 .go-contactless-benefit-card,
-.go-contactless-agency {
+.agency-card {
```
machikoyasuda, line 41: Really nitpicky but these aren't technically Cards. A Card has is an unit that has an image and text combined (https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Card). This repo does use Bootstrap, which has its own thing called Cards, so I don't think I would use the naming `card` here because it might make it seem like we are using Bootstrap Card stuff here. 

image

src/index.html
```diff
@@ -30,6 +30,50 @@ <h1 class="home-hero-title">A new home for modern and equitable transportation i
...
+        <div class="agency-card">
+          <img alt="Sacramento Regional Transit" src="/uploads/sacramento-regional-transit-sacrt.png" />
+        </div>
+        <div class="agency-card">
+          <img
+            alt="Santa Barbara Metropolitan Transit District"
+            src="/uploads/santa-barbara-metropolitan-transit-district-sb-mtd.png"
+          />
+        </div>
+      </div>
```
machikoyasuda, line 65: Since this segment of code is the same on both pages, it could be turned into an includes.

src/_sass/pages/_go-contactless.scss
angela-tran: Got it. I renamed it to just `agency` in https://github.com/cal-itp/mobility-marketplace/pull/491/commits/02bd720de6ffc057d17c16753328fab862d23b59

src/index.html
angela-tran: Extracted it to an includes in https://github.com/cal-itp/mobility-marketplace/pull/491/commits/121283eb94e705a0667509e610b1693c30088be7

machikoyasuda (approved):

Issue #440, opened 2023-04-13
SCOPE: Different implementation options for Provider Map


mrose914:
Given the [new redesign](https://www.figma.com/file/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=5601-29511&t=0sznFMu9eKaRxejn-0) of the Provider Map page, we would like to scope out implementation considerations and options for implementing the map, per convo in dev kickoff on 4/13.

segacy1:
@machikoyasuda @mrose914 I broke down the map functionality a little further [here](https://www.figma.com/file/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=5719-32260&t=R6VlS8K0U9poRjjo-4). The options included are:

1. (MVP) We keep the same functionality that exists today + the table
2. When a user clicks on a provider in the network list they'll see additional specs AND/OR
3. When a user clicks on a location on the map, the location is highlighted

PR #518, opened 2024-01-15
Automatically pull the Transit Agency dataset from the Cal-ITP warehouse

allejo:
This PR focuses almost exclusively on automating the fetching of Transit Agency (TA) provider data from the Cal-ITP data warehouse. It also updates the Provider Map page to fit the new data structure from the warehouse and some content updates.

Closes https://github.com/cal-itp/data-infra/issues/28

## Workflow Overview

- [Provider Map Jobs GHA Overview](https://github.com/cal-itp/mobility-marketplace/actions/workflows/provider-map-jobs.yml)
- [Provider Map Jobs Sample Run](https://github.com/cal-itp/mobility-marketplace/actions/runs/7534585728/job/20509198244)

## How the Worfklow works

1. The first thing this workflow takes care of is authenticating with Google with a dedicated service account, whose access key is stored in this repo's `GCP_SA_KEY` environment variable.
2. After we've authenticated, I setup and install Google Cloud SDK because we need the `bq` CLI tool to be able to download data from the warehouse
3. I use the `bq` tool to download the `mart_transit_database.dim_mobility_mart_providers` table as a CSV file.
4. Small caveat, due to [issue #315160970](https://issuetracker.google.com/issues/315160970), some warning messages end up polluting the CSV file, making it unparseable as CSV.
5. There is a workaround step with a `sed` command that deletes everything from the start of the file up until we find the name of the first column of the dataset (i.e. `agency_name`).
6. I then set up Python and Pandas, to do some quick data transformation and calculations in an embedded Python script.
7. I parse the `providers.csv` file that was downloaded from the warehouse and lookup to see what TAs have an empty `counties_served` column. I then use the `cities_to_county.csv` to have a lookup table of what counties a given city belongs to. This is necessary since the warehouse does not have `counties_served` and `hq_county` for every TA.

   I am working off the assumption that if a TA has its HQ in a specific city, that city can be assumed to be one the TA services.
8. After the data has been readied, I count every unique county across the `counties_served` column and update the `counties.geojson` file with that information.
9. Then I open a PR automatically with the **TWO** files I know should be touched in this workflow. See #513 for an example of this workflow succeeding

   Word of caution: do **NOT** blindly do a `git add .` (or its equivalent) in this PR. The `google-github-actions/auth` action will write your credentials to a file on the workflow's disk. If you accidentally add that, it'll cause you to expose the access key; which I did in #512 by accident.

## What's left to do

- [x] Los Angeles County (LAC) is an outlier in this dataset with 30 TAs and the next highest is Santa Barbara with only 5. This discrepancy leads to LAC being a super dark shade of purple and everything else is practically white.

   image

   ```
   Los Angeles       30
   Santa Barbara      5
   Ventura            3
   Riverside          3
   Alameda            2
   San Bernardino     2
   Tulare             2
   San Joaquin        2
   Kings              1
   Kern               1
   Placer             1
   Monterey           1
   Yolo               1
   San Mateo          1
   San Francisco      1
   Contra Costa       1
   Napa               1
   Santa Cruz         1
   Santa Clara        1
   Sacramento         1
   Sonoma             1
   ```
- [x] The provider map page is heavily broken because the new dataset is completely different in shape with the old one that was put together manually. Considering #440 plans for a redesign of the page, I've made removed any customizations that were used for the old dataset shape just so the new data can render (even if it looks ugly).

## What's next?

Not sure! /cc @thekaveman @machikoyasuda @angela-tran


allejo:
Since my last update to this PR, I've changed the `layers.json` for the state map in [`ddf54d4` (#518)](https://github.com/cal-itp/mobility-marketplace/pull/518/commits/ddf54d48fa7ef3caedf7cee3ace4c746717d2bf0#diff-2dab1e86b9104d41048f4ee9bb564ef5fe792a10209d37311a174112083b1ec9).

image

A few notes on my changes here. It appears the `5` represents now many different steps or shades there will be between the starting and ending colors. I changed the `0` from its original value of `1`; I'm not sure what this does but I know it allows for better shading between colors.

```json
"fill-color": [
  "interpolate",
  ["linear"],
  ["get", "num_providers"],
  0,
  "hsl(208, 36%, 78%)",
  5,
  "hsl(260, 53%, 36%)"
]
```

I also updated the column names so they appear on the Transit Agency table now.
	
---

The data dictionary is working, but it looks ugly. Not sure if you want to address that in this PR or in a future PR with a new design.

image

allejo:
> One follow up question, what mechanisms/controls do we have in place in the event the connection between the source data and the website goes down or is otherwise compromised? Want to make sure we have a way to detect issues with the view since we are now using it on the live site. And apologies if you already answered this - just need a little refresher :)

No apologies necessary! This is a good question that I haven't answered yet. This PR introduces a GitHub Action that runs on a schedule, every Sunday at midnight UTC (which translates to Saturdays 4pm). Should this process fail, we can trigger a Slack notification or something, depending on what is currently set up. @thekaveman would have a better idea on how we can set up notifications

evansiroky:
I'm a little concerned that this uses a scheduled GitHub action instead of it being a new DAG task within Airflow. Is there a reason we aren't using Airflow in this case?

evansiroky:
I'm concerned about some transit agencies not appearing in the preview site at https://deploy-preview-518--cal-itp-mobility-marketplace.netlify.app/ For example: `Burbank Bus`, `City of Fresno`, `SFMTA`.

allejo:
@mrose914 I've made the content updates you requested! and I changed how the Data Dictionary section displays, it's no longer a table
@evansiroky great catch! Turns out the way I was fetching the providers were limited to just 100 entries. I've updated it so that it returns everything and now Fresno is listed in there

---

Technical change, I've modified the CSV to JSON plugin so that it reads the CSV files into a `site.data.csv` variable. I am now generating the data dictionary statically at build time, which gives me better control of how to render it.

evansiroky:
Should I be seeing all of this data on the preview website? I'm only seeing one entry for Fresno County when there should be more.

image

allejo:
Oh man, sorry for all the back and forth. Okay, I _think_ this are displaying as they should be 🤞

image

evansiroky:
I think this looks good enough now. It's a little weird that City of Camarillo is showing up in Santa Clara County, but I think it's good enough for now.

evansiroky:
@mrose914 , I'm kind of just eyeballing and clicking around to check things, I'm not doing any kind of data comparisons.

What do you mean by copy? The changes I see here seem to be related to the list of providers that show up on the map.

evansiroky:
> @evansiroky got it, thanks.
> 
> Re: copy, see the headings in this comment: [#518 (comment)](https://github.com/cal-itp/mobility-marketplace/pull/518#issuecomment-1922482207)
> 
> There is some content describing the data on the page that needs to be updated to describe the latest data source(s). I could use some help in how we want to describe our data warehouse and this view to those who visit the page.

@mrose914 I see it now. Here is some proposed wording for this list of transit agencies:

*This dataset includes all publicly-funded transit agencies in California that provide fixed-route service that is available for the general public to ride without advance reservations.*

allejo:
Alright @mrose914, that should do it for the last copy edit! 🤞

.github/workflows/provider-map-jobs.yml
```diff
@@ -0,0 +1,111 @@
...
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+
+      - name: Install pandas
+        run: |
+          pip install pandas
+
+      - name: Update Counties GeoJSON file
```
thekaveman, line 52: I feel like this script would be easier to follow/debug/etc. if it was a standalone Python file, say in the `.github/workflows` directory next to this workflow.

Same goes for the requirements above, can we please have a `requirements.txt` that lists out `pandas` and pins a specific version -- and the workflow installs from there?

.github/workflows/provider-map-jobs.yml
```diff
@@ -0,0 +1,111 @@
...
+          with open(geojson_file, 'w') as f:
+            json.dump(geojson, f, indent=2)
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v5
+        with:
+          title: Provider Map Data Auto Update
+          body: |
+            It's that time again! The warehouse has delivered new data for us to use. This is an automatic pull request created by the `provider-map-jobs.yml` workflow; it is triggered via a cron that runs every Sunday at midnight UTC.
+          commit-message: Auto-update provider data from warehouse
```
thekaveman, line 107: Can we please have conventional commits formatting on this message to go with the general practice of our repos, e.g. `feat: ...` or `chore: ...` or even `feat(providers): ...` etc.

src/metadata/cities_to_county.csv
```diff
@@ -0,0 +1,456 @@
+﻿County,City
```
thekaveman, line 1: This is a nit, but I think this file would be easier to follow and diff any changes if the column order was `City,County` and/or if it was sorted by County first, then City.

thekaveman (changes requested):
Requesting some small changes to how the workflow is organized.

evansiroky:
> Thanks @allejo - copy is looking good. I'm inclined to remove the "Source" section at the bottom of the page given the content above and CTA to contact us for more info. @evansiroky Are you good with that or would you like to add something about how the list was developed?

Yep, fine with that. We can save the detailed dataset documentation for stuff we put on the [CA Open Data Portal](https://data.ca.gov/dataset/cal-itp-gtfs-ingest-pipeline-dataset/resource/e4ca5bd4-e9ce-40aa-a58a-3a6d78b042bd).

allejo:
@mrose914 I've updated the page to remove the source section ✅

_requirements.txt
thekaveman, line 1: Sorry to be a pain @allejo, but can you please move this into the `.github/resources` directory alongside the code? Don't want any confusion that these are requirements for the site itself.

.github/workflows/provider-map-jobs.yml
```diff
@@ -0,0 +1,63 @@
+name: Provider Map Jobs
+on:
+  # TODO: Remove this once behavior has been finalized
+  push:
+    branches:
+      - vlad/provider-map-gha
```
thekaveman, line 6: Just a reminder to remove this :grin:

thekaveman (approved):
Nice work!!! Thanks for sticking with this :smiley_cat:

thekaveman (approved):
:shipit: :rocket:

PR #524, opened 2024-02-07
fix(build): pin jekyll-sass-converter to fix build bug

### LINKED ISSUE #523: Bug: Cannot build jekyll locally
```
An error occurred while installing sass-embedded (1.70.0), and Bundler cannot
continue.

In Gemfile:
  jekyll was resolved to 4.3.3, which depends on
    jekyll-sass-converter was resolved to 3.0.0, which depends on
      sass-embedded
```

machikoyasuda:
fix #523 

Pinning `jekyll-sass-converter` fixes the bug for now. 

https://github.com/jekyll/jekyll-sass-converter/issues/161

I believe this issue is also affecting https://github.com/cal-itp/calitp.org


thekaveman (approved):
Confirmed this fixes the build issue on my (Windows) machine

PR #539, opened 2024-03-12
Add a better workaround so providers.csv isn't polluted

allejo:
An idiotic (but better than sed) workaround so that `bq` will not pollute our CSV file with a welcome message that is COMPLETELY ignored by `--quiet`, `--headless`, and every other flag that implies "silence unnecessary output."

see https://stackoverflow.com/q/73177304

### DIFF
--- a/.github/resources/download_csv.sh
+++ b/.github/resources/download_csv.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+bq query \
+    --quiet \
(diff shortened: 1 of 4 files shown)


thekaveman (approved):
Sheesh, more than silly that a CLI tool like this ignores flags telling it to be quiet and not spit out useless garbage :roll_eyes:

PR #543, opened 2024-04-01
Sort provider map query by NTD ID

allejo:
Looking at PRs #538, #541, #513, and gang shows no rhyme or reason behind the order of the CSV file. This makes it annoying and impossible to see the diffs between each pull, so this PR sorts the providers by NTD ID.

**Note:** PR #541 has been corrected to be in sorted order


thekaveman (approved):
Great change :+1:

PR #575, opened 2024-08-08
fix(nav): the for-vendors link can be set to current

### LINKED ISSUE #573: Bug: For Vendors page doesn't get highlighted when on page
image

machikoyasuda:
fixes #573 

There was a small bug in the menu code that was preventing the For Vendors link from being highlighted when you're on that page. The menu should highlight the current page you are on.

## Branch Preview vs. Prod
Branch Preview https://deploy-preview-575--cal-itp-mobility-marketplace.netlify.app/for-vendors
image

Production
image

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -24,7 +24,7 @@ navigation:
     - name: Contracts
       url: /contracts/
     - name: For vendors
-      url: /for-vendors/
+      url: /for-vendors
     - name: About
       url: /about
     - name: Contact us


angela-tran (approved):
Nice fix and debugging! I verified it works as expected on the preview site.

I dug into why it's ok for `/contracts/` to have a trailing slash: it's because [`contracts` is a folder](https://github.com/cal-itp/mobility-marketplace/tree/6e3e693b19f4120e254286e0adc53ae58a538182/src/contracts) with an `index.html` file, and either Jekyll or Netlify is making it so there is a trailing slash in the URL when you go to the Contracts page. 

The other pages are just an `.html` file, and [this logic](https://github.com/cal-itp/mobility-marketplace/blob/6e3e693b19f4120e254286e0adc53ae58a538182/src/_includes/link-repeater.html#L2) replaces the `.html` part (no trailing slash appended or anything).

PR #577, opened 2024-08-14
Improve SEO

allejo:
1. Add the Jekyll sitemap plugin to generate a sitemap that Google can use to index

   This won't fix all of our indexing issues because some of the issues are related to non-existent pages, error pages (e.g. 404, 500), or pages that have different URLs but have a canonical URL
2. Add OpenGraph descriptions to pages so that they can appear more helpful when being shared and when being indexed by Google

## Todo

- [x] Double check I've got all the pages updated with descriptions


src/contracts/index.html
```diff
@@ -1,6 +1,14 @@
...
 layout: skinny
+
 title: California’s Contracts—for U.S. transit agencies
+description: >
+  California has Master Service Agreements (MSA) available to help you modernize
+  your transit system, most of which are available to public transportation
+  providers. The Mobility Marketplace serves as Cal-ITP’s online catalog where
+  transit providers can browse available goods and services to build a system
+  that suits their needs best.
+
```
machikoyasuda, line 11: image
@mrose914 FYI - this is how these descriptions show up

machikoyasuda (approved):

PR #711, opened 2025-10-31
Ready for Ruby 3.5 and mise upgrade

### LINKED ISSUE #701: Ruby 3.5.0 upgrade + Netlify deprecations
These are known issues that need to be addressed sooner rather than later, to avoid accidental breakage in builds:

- [ ] jekyll-4.4.1/lib/jekyll.rb:26: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
- [ ] Explicitly enable idiomatic version files on Netlify
	```
	mise WARN  deprecated [idiomatic_version_file_enable_tools]:
	Idiomatic version files like /opt/build/repo/.ruby-version are currently enabled by default. However, this will change in mise 2025.10.0 to instead default to disabled.
	You can remove this warning by explicitly enabling idiomatic version files for ruby with:
	    mise settings add idiomatic_version_file_enable_tools ruby
	You can disable idiomatic version files with:
	    mise settings add idiomatic_version_file_enable_tools "[]"
	See https://github.com/jdx/mise/discussions/4345 for more information.
	```

allejo:
The `logger` Gem was originally part of the Ruby standard library, but in the upcoming 3.5.0 release, it will no longer be a part of the STD, so we have to include it explicitly in our Gemfile.

As for the Mise warning, it appears that in the next major release of Mise, it will no longer automatically install Ruby (or any other tool) if it detects files like `.ruby-version`. It sounds like this was changed to reduce the unintentional overhead that could occur if only one language needs to be installed; this mainly affects projects that use multiple. We now need to explicitly request that in a `.toml` file

Lastly, I updated the build image to Ubuntu 24.04 since our 20.04 image was deprecated. This was a config change in the Netlify panel.

Fixes #701


thekaveman (approved):
✅ The Netlify build preview looks good
✅ I was able to build and run the site locally in the devcontainer

PR #746, opened 2025-12-18
refactor: Set up a fresh Eleventy devcontainer

Scotchester:
ℹ️ **PR is to base branch `staging`** ℹ️ 

---

Collab with @jgravois 

The first commit deletes all source files and Jekyll-specific config. The second commit modifies the dev container to set up a barebones installation of 11ty, using the README content as a demonstration.

### Testing instructions

1. On this branch, rebuild and reopen the devcontainer.
2. The 11ty site will start up automatically, so once the process completes, you can go right to http://localhost:8080 to see the README content rendered as an HTML page (without any styling).

### Netlify

We think that this will build and deploy on the staging preview branch once merged there, since we changed the build command in `netlify.toml`.


thekaveman (changes requested):
I have to say (of course not having the context of the conversation the two of you had to work on this):

This scope of this PR feels wildly larger than "Config the devcontainer for Eleventy".

A single commit that deletes _everything_ else? That is rough, certainly from a review perspective, let alone the history this will (not) leave.

Even ignoring all this, we talked yesterday about _not starting npm when the devcontainer opens_ and I'd like to see that amended.

.devcontainer/devcontainer.json
```diff
@@ -3,21 +3,30 @@
   "dockerComposeFile": "compose.yml",
   "service": "site",
   "workspaceFolder": "/usr/src/mobimart",
+  "remoteEnv": {
+    "MISE_TRUSTED_CONFIG_PATHS": "${containerEnv:PATH}:/usr/src/mobimart"
+  },
+  "postStartCommand": [
+    "/bin/bash",
+    ".devcontainer/postStart.sh"
```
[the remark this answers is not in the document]
Scotchester: Thanks for clarifying the philosophy on devcontainer use. I didn't understand that aspect of it when we discussed yesterday. I had thought the concern was solely not wanting it to break immediately on startup and thereby kill the devcontainer, and I thought that the `sleep infinity` thing satisfied that concern _(turns out after additional testing, no it does not!)_, but I have no issue aligning to the philosophy you described. I'll remove the auto start and update `tasks.json` to be able to run it.

.devcontainer/Dockerfile
```diff
@@ -1,19 +1,23 @@
...
-RUN apt-get update
+# Install Mise for the same tool management as Netlify
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+ENV MISE_DATA_DIR="/mise"
+ENV MISE_CONFIG_DIR="/mise"
+ENV MISE_CACHE_DIR="/mise/cache"
+ENV MISE_INSTALL_PATH="/usr/local/bin/mise"
+ENV PATH="/mise/shims:$PATH"
+COPY .mise.toml /etc/mise/config.toml
+RUN curl https://mise.run | sh
```
jgravois: we tried a few different techniques to lay down Node.js and Python side by side. i'm fond of mise because it allows for managing the version of terraform, python and other tools that we use centrally.

one other benefit is that its what netlify uses under the hood too.

that said, I wouldn't be fussed to use `nvm` or something else entirely instead.

.devcontainer/Dockerfile
```diff
@@ -1,19 +1,23 @@
-FROM python:3.10
+FROM python:3.14
```
thekaveman: I appreciate the work to get `python` and `node` to play nicely together.

I'm not totally opposed to moving to `python:3.14` but will just point out we are standardized on `python:3.12` everywhere else.

Maybe this is a good opportunity to tease out any meaningful differences for our dev environment at least.

jgravois:
a couple more notes for posterity:

in order to get netlify to deploy this branch we had to clear the (i think redundant/unused) `RUBY_VERSION` environment variable set in the netlify UX to `3.1` (for branch previews specifically). the logs indicate that the .ruby-version file `3.4.5` was/is taking [precedence](https://app.netlify.com/projects/cal-itp-mobility-marketplace/deploys/69399d819b4d8700086ff2b7#L18) anyway.

we also noticed a couple weird but seemingly harmless individual quirks in our local containerized environments:
1. i had to `install 'code' command in PATH` using the VS Code command pallete. (scott did not)
1. scott had to `git config --global core.editor "code" --wait` (i did not)

PR #754, opened 2026-01-05
feat: Add step list component

Scotchester:

---

This PR adds the basic version of our styled list for a process with multiple steps (i.e., the condensed version that appears at the top of our guides, not the big list with the details of each step), which I am calling a "step list". Open to naming suggestions.

Example step list rendered with purple color scheme

In contrast to the pull quote component, I elected not to implement this as an include, because it seemed too gnarly / not beneficial enough when you'd have to pass in a full slate of `<li>`s in raw HTML, anyway. Open to disagreement on that, though!

The second commit makes sure that both the step list and pull quote components will render usefully if a `_color` modifier class is omitted from the markup by giving it a default light gray color scheme.

### DIFF
--- a/src/demo.md
+++ b/src/demo.md
@@ -47,9 +47,9 @@ description: >
 <div class="container">
   <h2>Step list</h2>
   <div class="row">
-    <h3>Default color scheme (yellow)</h3>
+    <h3>Color scheme: yellow</h3>
     <div class="offset-lg-2 col-lg-6 col-12">
-      <ol class="step-list my-4">
+      <ol class="step-list step-list_yellow my-4">
         <li><b>Choose a data-only provider</b> and plan that fits your agency’s needs</li>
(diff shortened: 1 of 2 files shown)


src/demo.md
```diff
@@ -43,3 +43,55 @@ description: >
...
 {% assign quote_source_title = "Senior Transit Planner, ICTC" %}
 {% include pull-quote.html color_scheme=quote_color_scheme content=quote_content source_name=quote_source_name source_title=quote_source_title %}
+
+<div class="container">
+  <h2>Step list</h2>
+  <div class="row">
+    <h3>Color scheme: yellow</h3>
+    <div class="offset-lg-2 col-lg-6 col-12">
+      <ol class="step-list my-4">
+        <li><b>Choose a data-only provider</b> and plan that fits your agency’s needs</li>
```
jgravois: TIL `<b>` is semantic again!

https://stellify.net/html5-b-and-i-tags-are-going-to-be-useful-read-semantic-again/ (from 2008 haha)

PR #756, opened 2026-01-05
feat: Add ability to attach a big number to any heading

Scotchester:
ℹ️ **PR is to base branch `staging`** ℹ️ 

---

Technically, they could be attached to any block element, but the positioning only works right if the element has a line-height of 1.25 (like our H2 and H3 styles). We can make that more robust later, if we encounter a need for it.

This borrows a lot from the styles set up for the step list in #754.

examples of numbered headings in all four color schemes

All it takes to add a number is to add a class of `numbered` and a `data-number` attribute to specify the value, along with an optional color scheme class (e.g., `numbered_yellow`):

```html
<h3 class="numbered numbered_yellow" data-number="1">Choose a data plan</h3>
```

### DIFF
--- a/src/styles/latest.css
+++ b/src/styles/latest.css
@@ -63,64 +63,90 @@
 /* STEP LIST */
 
 .step-list {
-  padding-left: 1.875rem;
-  padding-bottom: 1.5rem;
-  border-left: 0.1875rem solid var(--dsdl-gray-20);
-  margin-left: 0.9375rem;
+  /* This dance of padding/border/margin is needed to
+     align the left border with the center of the markers
+     and have the text end up 1rem from the marker.
+     The amounts total up to 48px / 3rem. */
+  padding-left: calc(30rem / 16);
+  border-left: calc(3rem / 16) solid var(--dsdl-gray-20);
+  margin-left: calc(15rem / 16);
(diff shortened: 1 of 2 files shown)


src/styles/latest.css
```diff
@@ -65,55 +65,85 @@
...
-  line-height: 2.0625rem;
   text-align: center;
 }
+.step-list li::before {
+  content: counter(step-list);
+  width: var(--spacing-4);
+  height: var(--spacing-4);
+  left: calc(-1 * var(--spacing-6));
+  top: -0.0625rem;
+  line-height: 2.0625rem;
```
jgravois: these values feel a little opaque to me. maybe this would be more expressive/legible/maintainable?

```suggestion
  top: calc(-1rem/16);
  line-height: calc(1rem/16 + 2rem);
```
if you disagree, not a hill i'd die on...

jgravois (approved):
CIL for an optional nit

src/styles/latest.css
jgravois: 😄 less is more!

Scotchester:
Cleaned up units and rebased!

Scotchester:
One thing I forgot to mention in the original PR description: `em` units are used here to ensure that the number will scale with whatever the heading size is.

PR #758, opened 2026-01-06
feat: Add table styling and examples

Scotchester:
Notes:
- Tables default to 100% width (of their container), but will stop shrinking once each column is only as wide as needed for the longest word in that column. At that point, they will overflow the container.
- Bootstrap's responsive table feature enables scrolling when a table overflows its container. - Because certain tables can get too squished horizontally for a readable experience, I added a `.wider` class that can force tables to be wider when their container is below 720px wide. (see the last example on the demo page)
- I have a couple questions in Figma for @cmajel, but they don't necessarily block getting this PR in.

### DIFF
--- a/src/styles/latest.css
+++ b/src/styles/latest.css
@@ -151,6 +151,59 @@
   color: white;
 }
 
+/* TABLES */
+
+table {
+  width: 100%;
+  border: 1px solid var(--dsdl-gray-40);
+  border-collapse: separate;
+  border-radius: var(--spacing-05);
+  border-spacing: 0;
+}
+/* Apply a border to the right of all but the last column */
+th:not(:last-child),
+td:not(:last-child) {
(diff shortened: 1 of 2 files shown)


src/styles/latest.css
```diff
@@ -151,6 +151,59 @@
...
+.table_cyan th {
+  background-color: var(--dsdl-cyan-10);
+}
+.table-responsive {
+  container-name: responsive-table-wrapper;
+  container-type: inline-size;
+}
+
+/* Magic numbers decided based on current table content. */
+@container responsive-table-wrapper (max-width: calc(720rem / 16)) {  /* Variables can't be used here yet. */
```
jgravois: i would have also accepted:
```suggestion
@container responsive-table-wrapper (max-width: 45rem) {  /* Variables can't be used here yet. */
```

jgravois (approved):
looking very sharp!

Issue #721, opened 2025-12-05
Design refinement: Color, text sizing, and link state adjustments


cmajel:
In reviewing the design files for dev handoff, there are a few places that need colors and button/link state adjustments to improve accessibility:

- [x] Review and adjust elements with white text on `cyan-40` backgrounds, which is low contrast (navigation, before and after buttons, quote styling)
- [x] Review and adjust button states for low contrast elements 
- [x] Review and adjust form inputs for low contrast elements 
- [x] Review mobile typescale
- [x] Adjust typography to implement headings/body copy updates noted in https://github.com/cal-itp/calitp.org/pull/547#pullrequestreview-3568381525

## Definition of done:
* Figma file is updated
* Review with dev team

PR #759, opened 2026-01-06
feat: Adjust type scale

Scotchester:
Follow up to #721

Makes adjustments to type scale per @cmajel's [new specs in Figma](https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=10714-5867&m=dev).

Specifically:
- H3 and H4: weight is bumped from semi-bold to bold
- H5: weight is reduced from semi-bold to medium and size is set at 18px on all screen sizes
- Desktop widths (>= 992px): Font size bumped up one increment for H1, H2, H3, H4, and small/caption text

If these adjustments prove successful, they should be relocated to the DSDL styles in calitp.org.

### DIFF
--- a/src/styles/latest.css
+++ b/src/styles/latest.css
@@ -1,3 +1,43 @@
+/* TYPE SCALE */
+
+/* If these adjustments prove successful, they should be relocated to the DSDL styles in calitp.org. */
+
+h3,
+.h3,
+h4,
+.h4 {
+  font-weight: 700;
+}
+h5,
+.h5 {
+  font-size: calc(18 / var(--text-base) * 1rem);
+  font-weight: 500;
+}
+
+@media (min-width: 992px) {
+  h1,
+  .h1 {
+    font-size: calc(48 / var(--text-base) * 1rem);
+  }
+  h2,
+  .h2 {
+    font-size: var(--text-xxxl);
+  }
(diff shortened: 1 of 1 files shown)


jgravois (approved):
i'm okay with this, but my preference would be to:
1. push this change to the `feature/dsdl` branch of calitp.org instead
1. use https://feature-dsdl--cal-itp-website.netlify.app/stylesheets/dsdl/dsdl.css in MobiMart to test this (and subsequent) tweaks
1. open a PR to calitp.org `main` once we're happy

Scotchester:
> i'm okay with this, but my preference would be to:
> 
>     1. push this change to the `feature/dsdl` branch of calitp.org instead
> 
>     2. use https://feature-dsdl--cal-itp-website.netlify.app/stylesheets/dsdl/dsdl.css in MobiMart to test this (and subsequent) tweaks
> 
>     3. open a PR to calitp.org `main` once we're happy

I like the idea conceptually, but properly integrating this stack with what's in the DSDL will require some additional consideration, I think. It doesn't harmonize neatly with the matrix of 12 core styles (`[Headline | Title | Subtitle | Body]` `[S | M | L]`). For that reason, I prefer to merge it this way for now.

PR #761, opened 2026-01-06
feat: Add download card component

Scotchester:
ℹ️ **PR base branch is `staging`** ℹ️ 

---

Minor deviation from the design: The web version of Space Grotesk served by Google Fonts does not include the download glyph, so it's falling back to the system font. This looks fine to me on macOS, but if folks prefer, we could get an SVG of the Space Grotesk glyph and use that instead.

row of three download cards

### DIFF
--- a/src/_includes/dl-card.html
+++ b/src/_includes/dl-card.html
@@ -30,7 +30,12 @@
     <p class="mb-3">{{ include.description | default: "" }}</p>
     <div class="dl-card-buttons">
       <a class="dl-button" href="{{ include.view_url }}">View <span class="visually-hidden">{{ include.title }}</span></a>
-      <a class="dl-button" href="{{ include.download_url }}" download>Download <span class="visually-hidden">{{ include.title }}</span> ⤓</a>
+      <a class="dl-button" href="{{ include.download_url }}" download>
+        Download <span class="visually-hidden">{{ include.title }}</span>
+        <svg viewBox="0 0 8 17" xmlns="http://www.w3.org/2000/svg">
+          <path d="M7.01999 16.38H-9.76026e-06V15.102H3.25799C2.55599 13.716 1.47599 12.402 -9.76026e-06 11.142V10.098C1.25999 10.692 2.23199 11.268 2.86199 11.826V-6.79493e-06H4.13999V11.844C4.76999 11.268 5.74199 10.692 7.01999 10.098V11.142C5.54399 12.402 4.46399 13.716 3.76199 15.102H7.01999V16.38Z" fill="currentcolor"/>
+        </svg>
+      </a>
     </div>
   </div>
 </div>
--- a/src/styles/latest.css
+++ b/src/styles/latest.css
@@ -334,6 +334,12 @@ td ol {
   background-color: var(--dsdl-cyan-80);
   color: white;
 }
+.dl-button > svg {
+  height: 1em;
+  margin-left: 0.25em;
+  position: relative;
+  top: calc(-1.5em / 18);  /* adjustment for optical alignment with the text */
+}
 .dl-button + .dl-button {
   margin-left: var(--spacing-05);
 }


src/styles/latest.css
```diff
@@ -256,6 +256,80 @@ td ol {
   }
 }
 
+/* DOWNLOAD CARDS */
+
+.dl-card {
+  display: block;
+  max-width: 332px;
+  aspect-ratio: 2 / 3;
```
jgravois: i've never seen this sizing utility used before. pretty cool!

src/styles/latest.css
```diff
@@ -256,6 +256,80 @@ td ol {
   }
 }
 
+/* DOWNLOAD CARDS */
+
+.dl-card {
+  display: block;
+  max-width: 332px;
```
jgravois: is this one we want to define in `px` rather than `rem` as a multiplier on our bsae spacing?

jgravois (commented):
forgive the premature review. this is already looking sharp!

src/styles/latest.css
Scotchester: Mmm, probably no reason for it not to be `rem`s. I just dropped it in there lazily while futzing with the sizing and forgot to convert it later.

src/styles/latest.css
```diff
@@ -256,6 +256,80 @@ td ol {
...
+  border: 1px solid var(--dsdl-gray-50);
+  border-radius: var(--spacing-1);
+  margin-bottom: var(--spacing-2);
+  position: relative;
+  background-color: var(--dsdl-cyan-10);
+}
+.dl-card-thumbnail {
+  display: block;
+  width: 100%;
+  box-shadow: 0 -2px 16px rgb(38 38 38 / 0.25);  /* gray 80 at 20% opacity */
```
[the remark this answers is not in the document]
jgravois: i had a hunch that it might have been a deliberate choice. i'm good either way. 🤙

jgravois (approved):
very snazzy. CIL for a few non-substantive nits.

if you want @cmajel's ⬇️ blessing before you merge this, no offense taken.

src/_includes/dl-card.html
```diff
@@ -0,0 +1,36 @@
...
+        * tag (string)          - text for the tag indicating the download category
+        * title (string)        - the download's title
+        * view_url (string)     - URL to the download's detail page
+        * download_url (string) - URL to the download itself
+
+    Optional Parameters:
+        * description (string) : '' - an additional short description for the download
+
+    Output:
+        Markup for a pull quote with optional attribution using the Bootstrap layout grid.
```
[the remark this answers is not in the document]
Scotchester: Ack, thanks.

src/_includes/dl-card.html
```diff
@@ -0,0 +1,36 @@
...
+{% endcomment %}
+
+<div class="dl-card">
+  <img class="dl-card-thumbnail" src="{{ include.img | default: "https://placehold.co/272x352/white/black?text=Scope+of+Work+template\n\n" }}" alt="">
```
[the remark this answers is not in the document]
Scotchester: Heh, I actually didn't mean to leave that in. I had tried to use the newlines as a lo-fi way to push the text up from the center, but at the end of the text the just get dropped, and in the middle of the text multiples get reduced to just one. That makes this particular placeholder service a nonstarter to me. Could look for another, but in practice I don't think we'll ever want to let one of those go to prod, so I think we might as well let it break more visibly so that it calls attention to itself when previewing.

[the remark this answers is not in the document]
jgravois: > I think we might as well let it break more visibly

that works for me too.

src/styles/latest.css
```diff
@@ -49,6 +49,12 @@ h5,
   border: none;
 }
 
+span.visually-hidden {
+  /* Deals with an issue that causes screen readers to read hidden span elements out of order:
+     https://github.com/h5bp/main.css/issues/12 */
+  margin: auto !important;
+}
+
```
jgravois: clever. i'm assuming that inverting the items in the DOM itself 👇 doesn't fix this all on its own, but if it did, i'd call that (slightly) tidier.

```html
<a class="dl-button" href="{{ include.view_url }}"><span class="visually-hidden">{{ include.title }} </span>View</a>
```

jgravois (approved):
getting your branch in sync with `staging` _should_ fix the deploy error.

src/styles/latest.css
Scotchester: The order in the markup is the order I _want_ it to read, but it's actually (before my fix) being read in the opposite order.

PR #770, opened 2026-01-13
feat: Add related nav component

Scotchester:

---

Adds a simple component for displaying the "Related" navigation (previous and next page) that goes at the bottom of each page.

Like the comparison block (#768), these currently fill the width of their container, so they look wider in the max-width preview than they will be once we have a narrower container in place.

These use the same hover/focus/active states as the global nav buttons and download buttons.

Comparing the related nav component across various viewports

### DIFF
--- a/src/styles/latest.css
+++ b/src/styles/latest.css
@@ -21,6 +21,35 @@ span.visually-hidden {
   margin: auto !important;
 }
 
+/* VARIABLES */
+
+:root {
+  --button-bg-color: white;
+  --button-border: 1px solid var(--dsdl-gray-50);
+  --button-border-radius: 2em;  /* enough for the ends to be perfect semicircles */
+  --button-font-size: calc(18 / var(--text-base) * 1rem);
+  --button-font-weight: 500;
+  --button-text-color: var(--dsdl-black);
+
+  --button-hover-bg-color: var(--dsdl-cyan-10);
+  --button-hover-border-color: var(--dsdl-cyan-60);
+  --button-hover-text-color: inherit;
+
+  --button-focus-outline: 3px solid var(--calitp-brand-blue);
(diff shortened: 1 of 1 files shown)


src/styles/latest.css
```diff
@@ -387,6 +387,40 @@ td ol {
...
+  text-decoration: none;
+}
+.related-nav-link:focus {
+  outline: 3px solid var(--calitp-brand-blue);
+  outline-offset: 2px;
+}
+.related-nav-link:hover {
+  border-color: var(--dsdl-cyan-60);
+  background-color: var(--dsdl-cyan-10);
+}
```
jgravois: if this declaration was centrally located it would be easier to avoid accidental divergence later if/when we decide to tweak.

```css
.global-nav a:focus,
.header a.contact:focus,
.related-nav-link:focus {
  outline: 3px solid var(--calitp-brand-blue);
  outline-offset: 2px;
}
```

if that isn't sufficient justification to lose the centralization of the styles for the component itself, we could also go the route of declaring variables for the properties themselves (since we can't use [`@mixins`](https://sass-lang.com/documentation/at-rules/mixin/)).

```css
: root { 
  --focus-outline: 3px solid var(--calitp-brand-blue);
  --focus-offset: 2px
}
```

src/styles/latest.css
```diff
@@ -387,6 +387,40 @@ td ol {
...
+.related-nav-link:active {
+  border-color: var(--dsdl-cyan-60);
+  background-color: var(--dsdl-cyan-60);
+  color: #fff;
+}
```
jgravois: at first i was thinking that we wouldn't ever be including a link to the current page, but this is 'active' in the sense of being clicked 🙃

PR #778, opened 2026-01-16
feat: /tap-to-pay

### LINKED ISSUE #732: Refactor: implement /tap-to-pay
figma: https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=8923-12154&m=dev

this is the one with the interactive fare calculator

jgravois:

resolves #732 

i tried reworking the history here in another branch to make it more obvious that we're actually selectively updating the content of /how-to/go-contactless.md but even with incremental commits for each step and preserving the markdown extension, the old file is interpreted as being deleted and the new one as added.

https://github.com/cal-itp/mobility-marketplace/compare/staging...feat/tap-to-pay-rebased?expand=1

### DIFF
--- a/src/tap-to-pay.html
+++ b/src/tap-to-pay.html
@@ -7,7 +7,7 @@ description: >
 
 <div class="container">
   <img width="100%" src="uploads/tap-to-pay-hero.png" alt="a hand holding a debit card to a contactless payment validator" />
-  <h1 class="text-center">Tap into the future of fare payment</h1>
+  <h1>Tap into the future of fare payment</h1>
   <p>
     California is making it easier for transit providers across the country to accept contactless credit, debit, mobile, or
     prepaid cards (including mobile wallets) through competitively awarded contracts.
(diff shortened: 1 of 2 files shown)


src/tap-to-pay.html
```diff
@@ -0,0 +1,474 @@
...
+  <div class="d-flex flex-md-row flex-column flex-wrap justify-content-center align-content-center gap-4">
+    {% for resource in site.data.resources -%} {% assign dl_img = resource.thumbnail %} {% assign dl_tag = resource.tag %} {%
+    assign dl_title = resource.title %} {% assign dl_description = resource.description %} {% assign dl_view_url =
+    "/resources/assets" %} {% assign dl_download_url = "/resources/" %} {% include dl-card.html img=dl_img tag=dl_tag
+    title=dl_title description=dl_description view_url=dl_view_url download_url=dl_download_url %} {% endfor %}
+  </div>
+</div>
+<div class="spacer-10"></div>
+<div class="container">
+  <h2 class="text-center">Related</h2>
```
Scotchester: Copying a comment I had accidentally left on an individual commit:

I was thinking we should set a global rule for H1 and H2 to be centered by default. What do you think?

jgravois: i think that makes sense 0f4f360

jgravois:
> did you try a single commit that does nothing but move/rename the file, and making the actual changes in other commits?

in [`#tap-to-pay-rebased`](https://github.com/cal-itp/mobility-marketplace/compare/staging...feat/tap-to-pay-rebased?expand=1) i laid down one commit to move the file, another to rename it and a third to update the content.

PR #808, opened 2026-01-28
Feat: update tap to pay

cmajel:
Adding copy updates to the tap to pay page.

Open threads on this copy:
- [x] Staff training materials (confirmed this PR is up to date / we're striking this for now)
- [x] Zip upload of the Sample User agreement docs (cc @jgravois )


src/tap-to-pay.html
```diff
@@ -373,14 +399,14 @@ <h4>Get ready to launch</h4>
   </ul>
   <p>
     Additionally, templated <a href="/TK">staff training guides</a> and
```
jgravois: we want to yank this link for now, correct?

regardless, i think the sentence should probably only mention 'staff training' once.

jgravois: rockin 🎸 but that sentence still duplicates "are available, are available" accidentally.

PR #824, opened 2026-02-02
chore: enable liquid-aware autoformatter (#823)

jgravois:
wires up [`shopify.theme-check-vscode`](https://marketplace.visualstudio.com/items?itemName=shopify.theme-check-vscode) for liquid aware autoformatting for .html files.

### DIFF
--- a/src/_includes/downloadable-resources.html
+++ b/src/_includes/downloadable-resources.html
@@ -22,8 +22,13 @@
         {% assign dl_view_url = '/resources/' | append: resource.slug %}
         {% assign dl_download_url = resource.filename %}
         {%
-          include dl-card.html img=dl_img tag=dl_tag title=dl_title
-          description=dl_description view_url=dl_view_url download_url=dl_download_url
+          include dl-card.html
+          img=dl_img
+          tag=dl_tag
+          title=dl_title
+          description=dl_description
+          view_url=dl_view_url
+          download_url=dl_download_url
         %}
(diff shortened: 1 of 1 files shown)


src/_includes/dl-card.html
Scotchester: I'm not sure why it isn't pulling back the indentation of all of the items in the opening comment (each line under Usage, each of the Params listed, etc.) to 2 spaces also, but I'm fine with that.

src/_includes/usage-estimator.html
```diff
@@ -3,17 +3,17 @@
...
-          <input
-            type="checkbox"
-            id="{{ category.id }}"
-            name="{{ category.id }}"
-            data-estimate="{{ category.estimate }}"
-            {{category.checked}}
-          />
-          <label for="{{ category.id }}">{{ category.label }}</label>
-        </div>
+        {% for category in site.data["usage-categories"] -%}
```
Scotchester: Huh, that's one I didn't expect, but it seems to work.

Scotchester (commented):
Aside from the remnants of the `"htmlWhitespaceSensitivity": "ignore"` experiment we discussed in Slack, I have one other request.

src/_includes/usage-estimator.html
jgravois: yeah. this is actually the preferred JS syntax.

PR #825, opened 2026-02-02
chore: yank unused hubspot form partial

jgravois:
this partial is inlined in 'contact.html' now:

https://github.com/cal-itp/mobility-marketplace/blob/475f7cbc024b8881ff45964b41c17e245e2bbff5/src/contact.html#L29-L33


machikoyasuda (approved):
💯 

Tested contact form, made sure the includes is not used anywhere in codebase.

PR #835, opened 2026-02-18
fix: ensure powerpoints and spreadsheets are passthrough copied

jgravois:
resolves https://cal-itp.slack.com/archives/C018XMWT2FN/p1771453380429109

this one was my mistake, when we migrated to 11ty i forgot to triple-check all the individual file extensions that we now need to whitelist.

while i was in here i noticed the description HTML we're trying to inject on the MSA Cost Estimation Tool preview page specifically results in some visual jank.

Screenshot 2026-02-18 at 2 53 34 PM

i pushed a commit with a stopgap fix for that too. i'll leave it for @machikoyasuda to think through something saner at her leisure.

* https://deploy-preview-835--cal-itp-mobility-marketplace.netlify.app/resources/msa-cost-estimation-tool/
* https://deploy-preview-835--cal-itp-mobility-marketplace.netlify.app/resources/contactless-payments-support-framework/

### DIFF
--- a/eleventy.config.js
+++ b/eleventy.config.js
@@ -16,6 +16,8 @@ export default async function (eleventyConfig) {
 
   eleventyConfig.addPassthroughCopy("src/uploads/**/*.pdf");
   eleventyConfig.addPassthroughCopy("src/uploads/resources/*.docx");
+  eleventyConfig.addPassthroughCopy("src/uploads/resources/*.pptx");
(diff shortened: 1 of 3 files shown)


Scotchester (approved):
Download works and description looks good on the preview deployment :+1:

jgravois:
i think that ultimately it was the right move to stop customizing the text size, but it turns out we could have also fixed this issue by switching to using single quotes in the declaration:

```html
<b class='text-body-xl'>Note:</b>

PR #837, opened 2026-02-20
Refactor: Refactor CSS into multiple files

### LINKED ISSUE #829: Refactor CSS into multiple files
For no reason other than inertia/time, we wrote all of the new MobiMart's CSS in a single file. Similar to what we did on `calitp.org` (https://github.com/cal-itp/calitp.org/pull/536), it'd be nice to break that up a bit so it's easier to find things.

We should also rename `latest.html` and `latest.css` so they are no longer relative to the work that is now complete :)

machikoyasuda:
closes #829 

## What this PR does

- Moves `variables`, `typography-forms`, `buttons`, `header`, `footer` into their own files. Imports these files into `main.css`. 
- Creates page-specific CSS files: `homepage`, `contact`, `404` - which are required in their respective files.
- Create layout-specific CSS files: `guides.css` - for all guide pages and `doc-preview` for document pages.
- Create CSS files for specific components: `related-nav`, `pull-quotes`, `download-cards`. The homepage, contact page CSS files import the components they use, like related nav and pull quotes.
- Rename latest.css/html to main.
- 11ty config: Edited the config to watch for all CSS files, following https://robmc.dev/blog/add-css-to-your-eleventy-site/ this blog post
 
## 📁  The resulting files
- `main.css` still has a few Bootstrap overrides and generic/shared/utility-type classes. At 50 lines, this file is much easier to parse and read.
- If, for whatever reason, a component used on only one layout/page is then going to be used somewhere else, the code can be pulled out and required in more than one style file.
- As much as possible, I tried not to modify any of the actual CSS. I did not delete the comments, for example, just to try to keep this PR as clean as possible.

### DIFF
--- a/src/data-plans.html
+++ b/src/data-plans.html
@@ -1,7 +1,7 @@
 ---
 layout: main
 title: Data plans
-stylesheet: guides.css
+stylesheet: guide.css
 description: We make it easy for transit providers to get the data plans they need. Learn if your agency can save on monthly connectivity costs.
 ---
 <div class="container">
--- a/src/gtfs-intro.html
+++ b/src/gtfs-intro.html
@@ -2,7 +2,7 @@
 layout: main
 
 title: Introduction to GTFS
-stylesheet: guides.css
+stylesheet: guide.css
 description: Everything you need to know about GTFS feeds and sharing real-time transit data.
 ---
 <div class="container">
--- a/src/gtfs-realtime.html
+++ b/src/gtfs-realtime.html
@@ -1,7 +1,7 @@
 ---
 layout: main
 title: GTFS Realtime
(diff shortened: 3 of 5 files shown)


src/styles/404.css
jgravois: to quote @Scotchester, it makes me a _tiny_ bit itchy to reference CSS variables like this that only resolve by virtue of the fact that `/main.css` happens to load first synchronously, but i'm on board with trying to avoid compiling the CSS for this site and i can't think of anything else superior off the top of my head, so this has my blessing.

Scotchester (changes requested):

jgravois (approved):
awesome! 🎸

PR #843, opened 2026-02-25
Fix: Remove MSA Cost Estimation tool description funkiness

### LINKED ISSUE #836: revisit MSA Cost Calculator description injection
in b2e75c1d036727304e77e0395a10718d254cf790 we started doing something gross to inject the MSA Cost Calculator description. (see #835 for more info)

if i had played a few more minutes i would have realized I could have just yanked the `text-body-xl` class on the `<b>` tag and left the HTML description in the front-matter like all the other pages.

@leelas-pm [thought that the difference in text size was a mistake](https://cal-itp.slack.com/archives/C018XMWT2FN/p1771455621556819?thread_ts=1771453380.429109&cid=C018XMWT2FN) so I _think_ everyone would be on board with having 'Note:' remain bold and instead just have the same font size as the surrounding text, but we should make sure.

while we're at it, i have a dim recollection that @cmajel felt this description in particular would benefit from being pared down. if that's true and we can spare the cycles, all the better.

https://www.camobilitymarketplace.org/resources/msa-cost-estimation-tool/

Image

machikoyasuda:
closes #836 

image

No more one-off logic in the layout.

### DIFF
--- a/src/_layouts/doc-preview.html
+++ b/src/_layouts/doc-preview.html
@@ -20,22 +20,6 @@ stylesheet: doc-preview.css
         <h1 class="mt-2 mb-lg-3 text-start">{{ resource.title }}</h1>
         <a class="dl-button my-3 my-lg-4" href="{{resource.filename}}" download>Download</a>
         <p>{{ description }}</p>
-        {% comment %}
-          see code comment in msa-cost-estimation-tool.md
-        {% endcomment %}
-        {% if page.fileSlug == 'msa-cost-estimation-tool' %}
-          <p>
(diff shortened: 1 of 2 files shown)


jgravois (commented):
if @cmajel is okay with the `Note:` being the same font size as it's surrounds, I think this looks great.

thanks for cleaning up my mess.

jgravois (approved):
🛁

jgravois:
just a breadcrumb to note something i just figured out this morning.

if we had truly needed to keep the CSS class on our `<b>` tag, we could have also got that working directly in the front-matter description by switching to single quotes.


```html
<b class='text-body-xl'>Note:</b>
instead of
<b class="text-body-xl">Note:</b>
```

PR #867, opened 2026-03-27
chore: replace homepage carousel with cards that link to guides

### LINKED ISSUE #865: Implement change from carousel to card layout for Home page guide treatment
See #786 for context. 

[Link to Figma for specs and finalized design. ](https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=8727-139&t=7KuyFJOIMSfNwAbC-0)

jgravois:
closes #865

protip: reviewing each commit in isolation leads to a significantly more legible diff imo.

before:
Screenshot 2026-03-30 at 2 46 18 PM
Screenshot 2026-03-30 at 2 46 26 PM

after:
Screenshot 2026-03-30 at 2 48 03 PM


src/index.html
```diff
@@ -61,94 +61,24 @@ <h2 class="h4 lh-1 fw-normal mb-0">Agencies we work with</h2>
...
-        <div class="carousel-caption">
-          {% comment %} <a class="h3" href="/rider-benefits">Explore rider benefits guide</a> {% endcomment %}
-          <div class="h3">Rider benefits</div>
-          <p class="text-body-xl mt-4">Offer tap to pay rider benefits from older adults to veterans to CalFresh and more</p>
+<section class="gallery-cards container-sm mt-5">
+  <h2 class="fw-normal">Guides to plan, purchase, and launch</h2>
+  <p class="gallery-description text-center">Our guides can help you leverage pre-approved MSAs and cut your procurement time.</p>
+  <ul class="gallery-cards row">
+    {% for card in galleryCards -%}
+      <li class="col-md-6 col-lg-6">
```
Scotchester: The overall width is wider than specified in Figma. The left edge should align with the left edge of the hero text (which is pushed over with `offset-lg-1`. I see two possible approaches:

1. Revise this loop so that the odd items are `<li class="col-md-6 col-lg-5 offset-lg-1">` and the even items are `<li class="col-md-6 col-lg-5">`
2. Just set a magic number max-width on `.gallery-cards` of 1090px.

src/styles/homepage.css
```diff
@@ -75,208 +75,82 @@
...
-.highlight > h3 {
+.gallery-card > h3 {
   color: var(--calitp-brand-blue);
   text-align: center;
 }
-.highlight > img {
+.gallery-card > img {
   display: block;
   width: 100%;
   margin: var(--spacing-4) 0;
```
Scotchester: Looks like we want this to now be margin-top only; there's too much space between the illustrations and the bottom of the cards.

```suggestion
  margin-top: var(--spacing-4);
```

src/styles/homepage.css
```diff
@@ -75,208 +75,82 @@
...
-  margin: 0;
-  gap: var(--spacing-2);
-}
-.carousel-indicators [data-bs-target],
-.carousel-controls > button {
-  box-sizing: initial;
-  width: auto;
-  height: auto;
-  padding: var(--spacing-1) 12px;
+.gallery-card a {
```
Scotchester: It looks to me like the styling for these buttons in Figma matches the styling of the download card buttons, but `.dl-button` is only available in `download-cards.css`, which is not imported on the homepage.

These are broken out as `button-small` [on the Components page](https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=11125-5898&m=dev).

Perhaps this is a good time to centralize that component in `buttons.css` and use it on both the download cards and these new guide cards?

src/styles/homepage.css
```diff
@@ -75,208 +75,82 @@
...
+.highlight {
+  padding: 24px 40px;
+  border: 1px solid var(--dsdl-gray-30);
+  border-radius: 40px;
+  margin-bottom: var(--spacing-2);
 }
-.guide-cards a.contact-cs:focus {
-  outline: var(--button-focus-outline);
-  outline-offset: var(--button-focus-outline-offset);
+.highlight > h3 {
```
Scotchester, line 128: Not a regression in this PR, but these headings shouldn't have top margin:

```suggestion
.highlight > h3 {
  margin-top: 0;
```

Scotchester (changes requested):
Looking pretty good! I identified a few deviations from the design that should be addressed before merging.

src/index.html
jgravois: good call. i went with option (1)

https://liquidjs.com/tags/cycle.html

src/styles/homepage.css
jgravois: this is a good suggestion too. i left the name `dl-button` for now, but i'm not opposed to renaming it in a follow-up PR. 457c9f5

src/_data/galleryCards.json
```diff
@@ -0,0 +1,34 @@
+[
+  {
+    "title": "Connect your fleet",
+    "description": "Save with pre-negotiated data plans for your fleets",
+    "href": "/data-plans/",
+    "hrefLabel": "Read the data plans guide",
```
Scotchester:
```suggestion
    "name": "data plans",
```

src/index.html
```diff
@@ -61,94 +61,24 @@ <h2 class="h4 lh-1 fw-normal mb-0">Agencies we work with</h2>
...
-          <p class="text-body-xl mt-4">Offer tap to pay rider benefits from older adults to veterans to CalFresh and more</p>
+<section class="gallery-cards container-sm mt-5">
+  <h2 class="fw-normal">Guides to plan, purchase, and launch</h2>
+  <p class="gallery-description text-center">Our guides can help you leverage pre-approved MSAs and cut your procurement time.</p>
+  <ul class="gallery-cards row">
+    {% for card in galleryCards -%}
+      <li class="col-md-6 col-lg-5 {% cycle 'offset-lg-1', '' %}">
+        <div class="gallery-card">
+          <h3 class="text-start mt-0">{{ card.title }}</h3>
+          <p>{{ card.description }}</p>
```
Scotchester: This text uses the "p2" size in Figma.

```suggestion
          <p class="text-body-xl">{{ card.description }}</p>
```

Scotchester (changes requested):
Thanks for addressing my earlier feedback. Couple more little nits on second pass.

Scotchester (dismissed):
Looks good to this dev, but would recommend that @cmajel take a look before we merge.

cmajel (commented):
Nice! I have a couple of design notes on this:

* We've got a couple H2 styles happening across sections. Can we switch this headline to the standard H2 style used in the following sections?
>Guides to plan, purchase, and launch

* This intro line should be a `p2` style at 20px font size (currently 16px)
> Our guides can help you leverage pre-approved MSAs and cut your procurement time. 

* There should be more spacing between the last card and the pull quote. Can we increase this to a total of 72px / `spacing-9`?

* When testing on mobile, I noticed I can do a horizontal scroll on this branch. It doesn't appear on the live site, so I assume it's something with the card implementation? Could you investigate?

Image

* A minor thing on mobile. I noticed when I click a card button and then go back to home, the card button returns with a hover state. Is that something we can reset and have it just show the default style?

Image

jgravois:
> I have a couple of design notes on this

great catches @cmajel. ddf3878 should address everything you noticed. 🙏

jgravois:
> A minor thing on mobile. I noticed when I click a card button and then go back to home, the card button returns with a hover state. Is that something we can reset and have it just show the default style?

this is one i left alone. that behavior is more noticeable on a mobile device but it's not actually peculiar to that form factor. on the desktop, the hover style persists too, (ie: it doesn't actually reset until you move your mouse over the page again.)

output

since its not something newly introduced in this PR my instinct is to track it in a separate issue or just let sleeping dogs lie.

cmajel:
@jgravois re the mobile hover issue it was more a curiosity as to why it was happening. I'm fine tracking it as a separate, low priority issue. 

One additional thing. It appears the card images aren't the updated Figma and my guess are using the past carousel images. I reformatted the layouts to better fit in the card. Can you update them? (adding files here)
[artbox-cards.zip](https://github.com/user-attachments/files/26442482/artbox-cards.zip)

jgravois:
> I'm fine tracking it as a separate, low priority issue.

i'm all for it. could be an easy fix.

> I reformatted the layouts to better fit in the card

ah, sorry i missed that. fixed in 2f8e3fe and enabled auto-merge.

@Scotchester i didn't run these images through ImageOptim because the ones I replaced were blurrier in comparison. but i'm happy to revisit this down the line if you have a mind to.

cmajel:
Updates look good! No more notes from me. 🚀  

Re ImageOptim adding a 2x export too if that is useful - ignore if it is not. 
[artbox-cards-2x.zip](https://github.com/user-attachments/files/26461262/artbox-cards-2x.zip)

jgravois:
thx @cmajel! can you give this one an official ✅?

@Scotchester won't be back on websites until next month.

cmajel (approved):
🚀 ✅

PR #869, opened 2026-03-31
Improve table responsiveness on small screens

### LINKED ISSUE #845: Revisit mobile table resizing/scrolling
On mobile, the tables extend beyond the content container and cause an unexpected horizontal scroll for the whole page (GTFS is the most noticeable). It appears the intent was for the table to scroll horizontally on mobile, with the left col staying sticky.

Image

Image

Scotchester:
Closes #845 

---

- Adds `.table-responsive` wrapper to tables in guide so that only they scroll horizontally, not the whole page
- Makes the left columns of responsive tables sticky when scrolling them horizontally

Side effects:
- Converted the first cell of body rows to a `<th>`, which seems semantically correct but changes the original design – they now have the same background color and bolding as column header cells.
- Removes the "wider" class, which was used on only one table (on the GTFS-RT page), because it made the untstuck columns too wide to fit between stuck column and edge of the scrolling container.
- Responsive tables no longer have rounded corners, because it resulted in odd visual artifacts when combined with the sticky column behavior.

Demo of table scrolling on a phone-size screen

### DIFF
--- a/src/data-plans.html
+++ b/src/data-plans.html
@@ -103,7 +103,7 @@ connectivity to modernize and improve the rider experience.' %}
   <h4 id="contracts" class="mt-3">Data plans MSAs</h4>
   <p>Cost per month per vehicle, prices are negotiable</p>
   <div class="table-responsive my-3">
-    <table class="table_purple mb-1">
+    <table class="table_purple mb-2">
       <thead>
         <tr>
           <th scope="row">MSAs</th>
--- a/src/data-plans.html
+++ b/src/data-plans.html
@@ -171,9 +171,6 @@ connectivity to modernize and improve the rider experience.' %}
           <td>–</td>
         </tr>
       </tbody>
-      <caption class="visually-hidden">
-        Data plans MSAs (cost per month per vehicle, prices are negotiable, click on each MSA for terms and conditions)
-      </caption>
     </table>
   </div>
   <p class="table-caption">Click on each MSA for terms and conditions.</p>
--- a/src/data-plans.html
+++ b/src/data-plans.html
@@ -102,78 +102,80 @@ connectivity to modernize and improve the rider experience.' %}
   </p>
   <h4 id="contracts" class="mt-3">Data plans MSAs</h4>
   <p>Cost per month per vehicle, prices are negotiable</p>
-  <table class="table_purple my-3">
-    <thead>
-      <tr>
-        <th scope="row">MSAs</th>
-        <th colspan="3" scope="col">
-          <a href="https://cdt.ca.gov/services/calnet-calitp/">CALNET</a> +
-          <a href="https://resources.calitp.org/mobility-marketplace/State.of.California_Amendment.1_Extension.2023-05.pdf"
-            >NASPO</a
-          >
-        </th>
-        <th scope="col">
-          <a href="https://resources.calitp.org/mobility-marketplace/State.of.California_Amendment.1_Extension.2023-05.pdf"
-            >NASPO</a
-          >
-        </th>
-      </tr>
-      <tr>
-        <th scope="row">Vendors</th>
-        <th scope="col">AT&T</th>
-        <th scope="col" class="text-nowrap">T-Mobile</th>
-        <th scope="col">Verizon</th>
-        <th scope="col">FirstNet</th>
-      </tr>
-    </thead>
-    <tbody>
-      <tr>
-        <th scope="row">1 GB</th>
-        <td>$20</td>
-        <td>$8</td>
-        <td>$15</td>
-        <td>$7.50</td>
-      </tr>
-      <tr>
-        <th scope="row">2 GB</th>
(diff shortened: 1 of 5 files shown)


src/tap-to-pay.html
```diff
@@ -133,232 +133,238 @@ <h4 class="mt-3">PADs (Payment Acceptance Devices) MSAs</h4>
...
+              <li>Mastercard</li>
+              <li>Visa</li>
+            </ul>
+          </td>
+        </tr>
+      </tbody>
+      <!-- are these captions redundant? -->
+      <caption class="visually-hidden">
+        Payment Processor MSAs (click on each MSA for terms and conditions)
+      </caption>
```
jgravois: neither here nor there in this PR, but i still suspect the value of these captions is dubious given the preceding `<h5>`.

happy to defer to @Scotchester. if they're helpful, lets ditch the code comments.

Scotchester: I've had the notion that tables always need a caption for accessibility in my head for a long time, but upon further review of reputable guidance, I think I agree that it's not necessary when we have the heading in close proximity, and our content is such that it's unlikely people would be navigating with a screen reader by jumping to tables. I'll go ahead and remove them here.

jgravois: its a tangent, so thanks for humoring me.

src/data-plans.html
```diff
@@ -102,78 +102,80 @@ <h3 class="numbered numbered_purple" data-number="2">Review purchase options</h3
...
-        <td>$34.99</td>
-        <td>–</td>
-      </tr>
-    </tbody>
-    <caption class="visually-hidden">
-      Data plans MSAs (cost per month per vehicle, prices are negotiable, click on each MSA for terms and conditions)
-    </caption>
-  </table>
+  <div class="table-responsive my-3">
+    <table class="table_purple mb-2">
```
jgravois, line 106: i'm not opposed to adding a wrapping div if its helpful, but something like this certainly feels more 'encapsulated'.

```css
table.table-responsive {
  /* ... */  
}
```

```html
<table class="table-responsive table_purple mb-2 my-3">
```
would that be problematic?

regardless, i'm assuming its a BEM thing to hyphenate `table-responsive` and underscore `table_purple`. (ie: the latter is a variant)

Scotchester: `table-responsive` comes from Bootstrap, which partially explains both of your points.

I think the wrapper is necessary because something has to serve as the container that fits within the page but can scroll horizontally when its contents (the actual table) overflow that.

As far as the color variant naming goes, I think I fell into this pattern because I started with the pull quote component, which already has a space in its name, so if it had a base class of `pull-quote` and I wanted to add a variant, it seemed intuitive to separate it with an underscore (e.g., `pull-quote_purple`) so that multiple hyphens didn't look ambiguous.

In hindsight, it may have been better to hew more closely to Bootstrap. I don't love that it doesn't really make any distinction between element and modifier classes (everything is just separated with hyphens, no matter what kind of class it is), but the consistency would probably be better. Fixing that is probably something for another day, though.

jgravois: > Fixing that is probably something for another day, though.

agree 100%.

thanks for the history lesson. i'll wait for @cmajel to chime in before i approve but this is looking great as far as i'm concerned.

cmajel:
@Scotchester ty for the clear write up on this.

I'm ok with these tradeoffs from the design side and didn't see any issues testing this on my desktop or phone. 🚀

src/styles/guide.css
```diff
@@ -155,14 +169,6 @@ td ol {
...
 }
 
-/* Magic numbers decided based on current table content. */
-@container responsive-table-wrapper (max-width: calc(720rem / 16)) {
-  /* Variables can't be used here yet. */
-  .wider {
-    min-width: calc(var(--spacing-base) * 120);
-  }
-}
-
```
jgravois, line 165: does this make the declaration below moot?

```css
.table-responsive {
  container-name: responsive-table-wrapper;
  container-type: inline-size;
}
```

src/styles/guide.css
```diff
@@ -118,6 +118,20 @@ table {
   border-radius: 4px;
   border-spacing: 0;
 }
+.table-responsive table {
+  border-color: var(--dsdl-gray-40);
+  border-style: solid;
+  /* left border is applied below to cells in first column so it stays in place when scrolling horizontally */
+  border-width: 1px 1px 1px 0;
+  border-radius: 0;
+}
```
jgravois, line 127: i was surprised to find that bootstrap only sets one useful prop on `.table-responsive`. 😇 

Image

jgravois (commented):
CIL for one nit. lookin good though.

src/styles/guide.css
Scotchester: Ah yeah, good catch!

jgravois (approved):
very snazzy ✨

Issue #875, opened 2026-04-15
Implement Rider Benefits page


angela-tran:
This covers adding all the content for the Rider Benefits page, except for the Downloadable Resources part.

Issue #873, opened 2026-04-14
Build and deploy Benefits page to MobiMart


leelas-pm:
[Link to Figma main branch with merged changes here. ](https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=12365-6183&t=3UFJhuGEjIraoiHz-0)

URL: https://www.camobilitymarketplace.org/rider-benefits/

jgravois:
i threw a few 'take em or leave em' comments in Figma, but overall i think the new page looks great! 🛹 

i'd like @angela-tran to run point on implementing this, but i expect it to be an exceedingly reasonable lift.

cmajel:
Adding a few nav notes here to complement Figma. This would also include adding:

- A new nav item "Rider benefits" to the top nav and footer nav
- Adding a "Rider benefits" card to the homepage (design is in Figma)

PR #878, opened 2026-04-15
chore: add hero and showcase images for rider benefits

jgravois:
adds the hero image we'll need for #875 and the showcase image we'll need for #874.

after downloading the high-res images from Figma I ran them through [ImageOptim](https://imageoptim.com/mac).

@cmajel if its okay with you, next time we tackle a redesign I'd love to make a point of optimizing the files there in the interest of maintaining a single source of truth. 🌈 

## Steps to test

this PR should not affect the live site. its just laying groundwork for #873.


angela-tran (approved):
Thanks @jgravois!

Issue #877, opened 2026-04-15
Update "Downloadable resources" section and tag logic to support multiple tags


angela-tran:
Currently, [/tap-to-pay](https://www.camobilitymarketplace.org/tap-to-pay/) surfaces related resources tagged as <kbd>Tap2Pay</kbd> and the [/gtfs-realtime](https://www.camobilitymarketplace.org/gtfs-realtime/) surfaces ones tagged as <kbd>GTFS-RT</kbd>

For the new /rider-benefits guide page we need to append an additional <kbd>Rider benefits</kbd> tag to 'Marketing Materials' and preserve its existing <kbd>Tap2Pay</kbd> tag.

To accomplish this we'll need to update the behavior of the app to:

* display all tags associated with a resource on its preview page
* display all tags associated with a resource in its preview card
* enhance the filtering logic for the downloadable resources component to filter on _all_ matching tags.

ie: the Marketing Materials doc preview page should now include resources tagged as either <kbd>Tap2Pay</kbd> _or_ <kbd>Rider benefits</kbd>

<img width="565" height="607" alt="Image" src="https://github.com/user-attachments/assets/e544524b-9bbd-4910-aa5c-28b5b79c1e46" />

cmajel:
Adding a recap of what we discussed:
* Add two tags to this resource that appear on the doc preview card and the doc page. ([In progress Figma branch with multiple tags for the marketing doc](https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/branch/6bTxuti5aokSLJ51hNNCyc/Mobility-Marketplace?node-id=9115-4674&t=LtTQ7KgUehIiE3yd-0) to explore this.))
* The other resource that falls into this pattern is the Standard Agreement (GTFS, Tap2pay), however we plan to retire this in #881. 

My instinct is to pair this work with #881. Let me know if that feels right, and I'll work that into this branch for simplicity.

cmajel:
@angela-tran @jgravois 

Here's the [Figma branch](https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/branch/6bTxuti5aokSLJ51hNNCyc/Mobility-Marketplace?node-id=9115-4674&t=hGL0ISGn4RJ74zKd-0) which includes:
* Patterns for multiple tags for Tap to Pay Marketing Materials
* Removing the Standard Agreement (#881), which is the other resource that currently has two references.

[...]

PR #886, opened 2026-04-17
feat: deprecate STD resource pages and update copy of remaining user agreements

### LINKED ISSUE #887: Chore: deprecate STD user agreement resources
as described in #881

jgravois:
resolves #887

I wired up a new markdown parser in order to support the embedded hyperlinks in the design so i opted to land this separately from the work to beef up the way we handle tags in #877.

* removes both STD resource pages
* updates the copy of the remaining user agreements to link out directly
* all other doc preview pages should look the same as they do in production
* add redirects for the removed pages


cmajel:
@jgravois redirects look great 🎉 

This was an oversight on my part, but can we add the external link indicator to the Standard agreement link? (preview below)

Screenshot 2026-04-20 at 9 24 03 AM

jgravois:
> can we add the external link indicator to the Standard agreement link?

great catch. done in 6259586

src/resources/contactless-payments-introduction.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: doc-preview
-description: >
+description: |
```
jgravois, line 3: this literal block scalar is only truly needed when the description contains multiple paragraphs, but it seemed most appropriate to me to use it consistently.

src/resources/user-agreement-gtfs.md
```diff
@@ -1,7 +1,9 @@
 ---
 layout: doc-preview
-description: >
-  A sample document, to be used alongside the GTFS Realtime Scope of Work (SOW) template and the corresponding email template. The User Agreement contract incorporates, by reference, the terms of the MSA and contains the Scope of Work and Payment Provisions. Please note that the User Agreement contract can also be the Standard Agreement (STD 213) provided by DGS.
+description: |
+  A sample document, to be used alongside the GTFS Realtime Scope of Work (SOW) template and the corresponding email template. The User Agreement contract incorporates, by reference, the terms of the MSA and contains the Scope of Work and Payment Provisions.
+
+  The User Agreement contract can also be the [Standard Agreement (STD 213) <span aria-label='(external link)'>↗</span>](https://www.dgs.ca.gov/ols/Forms?search=213&topicCategoryFilters=&audienceCategoryFilters=9048d318e266493abfd2361dc40d1f45&sort=relevance&activeFilters=audience%7C&page=1) provided by DGS. You may need Adobe Reader or Adobe Acrobat to view these documents.
```
jgravois, line 6: the reason i switched to markdown originally was to avoid the pesky rendering bug we first patched in #835

when @cmajel reminded me that we need an external link indicator, i had to take a second pass and enable HTML parsing again. thats when i realized that its not the HTML tags themselves that are problematic, but rather the inline `"` wrapped property declarations.

ie:
```html
<!-- this is parsed correctly -->
<span aria-label='(external link)'>↗</span>

<!-- this is not -->
<span aria-label="(external link)">↗</span>
```

its a bit of a footgun, but when folks have got it wrong, its readily apparent.

cmajel:
@jgravois ty for jumping on this quickly.

I noticed the Tap to Pay page has some wonk in the displayed text. Probably a syntax error. 
Screenshot 2026-04-20 at 1 15 34 PM

jgravois:
> the Tap to Pay page has some wonk in the displayed text

derp. thank you! f0c3405

angela-tran (approved):
Very thorough and straightforward 👍

PR #888, opened 2026-04-17
feat: add support for resources with multiple tags

### LINKED ISSUE #877: Update "Downloadable resources" section and tag logic to support multiple tags

jgravois:
resolves #877

## how to test

1. [/tap-to-pay](https://deploy-preview-888--cal-itp-mobility-marketplace.netlify.app/tap-to-pay) and [/gtfs-realtime](https://deploy-preview-888--cal-itp-mobility-marketplace.netlify.app/gtfs-realtime) display the same downloadable resources as before
1. individual doc preview pages show the same related resources as before
1. [/resources/marketing-materials](https://deploy-preview-888--cal-itp-mobility-marketplace.netlify.app/resources/marketing-materials/) displays an additional `Rider benefits` tag
1. [/demo](https://deploy-preview-888--cal-itp-mobility-marketplace.netlify.app/demo/) displays `GTFS-RT` resources (fixes a bug i found along the way)

### DIFF
--- a/src/_includes/downloadable-resources.html
+++ b/src/_includes/downloadable-resources.html
@@ -2,7 +2,7 @@
   Usage:
     {% include 'related-documents.html' resource_tag:'Tap2Pay' %}
   Parameters:
-    * resource_tag (string) - the name of a tag to filter on
+    * resource_tags (Array of strings) - the names of a tags to filter on
   Optional Parameters:
     * heading (string) : 'Downloadable resources' - the heading to display above the cards
     * current_slug (string) : to avoid displaying a card for the resource currently being previewed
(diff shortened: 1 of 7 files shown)


angela-tran (approved):
Looks good!

jgravois:
> Looks good!

thx @angela-tran. i just rebased and `-f` pushed to enable the <kbd>Merge pull request</kbd> 😇

src/_includes/downloadable-resources.html
```diff
@@ -2,7 +2,7 @@
   Usage:
     {% include 'related-documents.html' resource_tag:'Tap2Pay' %}
   Parameters:
-    * resource_tag (string) - the name of a tag to filter on
+    * resource_tags (Array of strings) - the names of a tags to filter on
```
angela-tran, line 5: Just noting this for future reference / to check my understanding: introducing this wasn't necessary for this go-around? -- that is, each ~page~ guide that uses this include template still only represents 1 tag, i.e. passes in 1 tag. But this makes it possible for a ~page~ guide to show resources for more than one tag

jgravois: thats 95% correct. the only caveat being that the doc-preview page relies on this include too:

https://github.com/cal-itp/mobility-marketplace/blob/da5b5a1090b70079d924b0003ffcaa99c785f78c/src/_layouts/doc-preview.html#L33-L37

since we only have one resource tagged as <kbd>Rider benefits</kbd> we could have also introduced a kludge to make sure that that page in particular showed related results for <kbd>Tap2Pay</kbd> specifically and got the same visual result.

i took this tack instead because it seemed just as easy (and more future-proof).

angela-tran: Ah, I see - thanks for pointing out that use case! I was focusing only on the one-item `resourceTags` in the frontmatter for the guides.

I see now with the doc-preview use case in mind that `resourceTags` was necessary -- it'll evaluate to `["Rider benefits", "Tap2Pay"]` for the "Tap to pay marketing materials" resource via https://github.com/cal-itp/mobility-marketplace/blob/da5b5a1090b70079d924b0003ffcaa99c785f78c/src/_layouts/doc-preview.html#L11

Definitely the better solution :+1: Thanks for the explanation

PR #879, opened 2026-04-15
Feat: Benefits product page

angela-tran:
Closes #873 

view it live: https://deploy-preview-879--cal-itp-mobility-marketplace.netlify.app/rider-benefits/

PR #891, opened 2026-04-21
Feat: Benefits downloadable resource and "Related" section

### LINKED ISSUE #890: Add Downloadable Resource and Related sections for Benefits page
### "Downloadable resources" section

Pre-requisite changes:
- #886 
- #888 

With those PRs done, the structure is in place to simply add the "Rider benefits" tag to the "Tap to pay marketing materials" resource.

### "Related" section

See https://github.com/cal-itp/mobility-marketplace/pull/891#issuecomment-4298651635

angela-tran:
Closes #890 

**Downloadable resources**
- Adds `resourceTags: ["Rider benefits"]` to `rider-benefits` frontmatter and adds usage of `downloadable-resources` include template to `rider-benefits`

**"Related" section**
- Adds the "Related" section to `rider-benefits`, and updates the next link on `tap-to-pay` to go to `/rider-benefits`
  - `rider-benefits`'s "previous" link goes to `/gtfs-realtime` and "next" link goes to `/tap-to-pay`
  - **Open question**: this implementation matches what Figma shows, but I'm wondering if the design intention was for these "previous" and "next" links to follow the ordering of the nav links, in which case `rider-benefits`'s "previous" link should be to `/tap-to-pay` and "next" link should be to `/data-plans`. `data-plans`'s "previous" link would be for `/rider-benefits`. 
  - **Open question**: is the styling on the "next" link something that we're introducing as a part of this Benefits page effort? Or was that a style in Figma that was never implemented, in which case I guess the options are to ignore it/remove it from Figma, just implement it, or leave it as something that we want to implement but not right now?

| Figma | Currently live MobiMart |
| --- | --- |
| image | image |

### DIFF
--- a/src/data-plans.html
+++ b/src/data-plans.html
@@ -316,9 +316,9 @@ connectivity to modernize and improve the rider experience.' %}
   <h2 class="mt-0 mb-3 pb-3">Related</h2>
   <nav class="related-nav row mb-4 pb-3 pb-0" aria-label="Related">
     <div class="col-md">
-      <a class="related-nav-link" href="/tap-to-pay">
+      <a class="related-nav-link" href="/rider-benefits">
         <span class="arrow" aria-label="(previous page)">←</span>
-        <span class="text">Learn more about tap to pay rider benefits</span>
+        <span class="text">Make your tap to pay program more robust with <b>rider benefits</b> & discounts</span>
       </a>
     </div>
     <div class="col-md">
--- a/src/tap-to-pay.html
+++ b/src/tap-to-pay.html
@@ -520,7 +520,7 @@ with contactless hardware and fare-calculation and payment-processing software.'
     </div>
     <div class="col-md">
       <a class="related-nav-link related-nav-link_next" href="/rider-benefits">
-        <span class="text">Make your tap to pay program more robust with rider benefits & discounts</span>
+        <span class="text">Make your tap to pay program more robust with <b>rider benefits</b> & discounts</span>
         <span class="arrow" aria-label="(next page)">→</span>
       </a>
     </div>
--- a/src/tap-to-pay.html
+++ b/src/tap-to-pay.html
@@ -519,8 +519,8 @@ with contactless hardware and fare-calculation and payment-processing software.'
       </a>
     </div>
     <div class="col-md">
-      <a class="related-nav-link related-nav-link_next" href="/data-plans">
-        <span class="text">Explore how to increase data or save on your data plans</span>
+      <a class="related-nav-link related-nav-link_next" href="/rider-benefits">
+        <span class="text">Make your tap to pay program more robust with rider benefits & discounts</span>
         <span class="arrow" aria-label="(next page)">→</span>
       </a>
     </div>
--- a/src/gtfs-realtime.html
+++ b/src/gtfs-realtime.html
@@ -319,7 +319,7 @@ description: Follow this step-by-step guide to launch a GTFS Realtime program an
     </div>
     <div class="col-md">
       <a class="related-nav-link related-nav-link_next" href="/tap-to-pay">
-        <span class="text">Explore tap to pay to bring contactless payments to your fleet</span>
+        <span class="text">Explore <b>tap to pay</b> to bring contactless payments to your fleet</span>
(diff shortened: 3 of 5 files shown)


angela-tran:
Hi @cmajel - there are 2 open questions up above about the "Related" section for the Benefits page. You can see how it's currently working with https://deploy-preview-891--cal-itp-mobility-marketplace.netlify.app/rider-benefits/

@jgravois - do you happen to know why there's drift on the "next" link style between Figma and the live site?

src/tap-to-pay.html
```diff
@@ -519,8 +519,8 @@ <h2 class="mt-0 mb-3 pb-3">Related</h2>
       </a>
     </div>
     <div class="col-md">
-      <a class="related-nav-link related-nav-link_next" href="/data-plans">
-        <span class="text">Explore how to increase data or save on your data plans</span>
+      <a class="related-nav-link related-nav-link_next" href="/rider-benefits">
+        <span class="text">Make your tap to pay program more robust with rider benefits & discounts</span>
```
jgravois: it makes sense to me to emphasize the name of the guide in the text as we do in the other links, but lets hold off on this until we ask for a formal round of QA from outside the dev team.

```suggestion
        <span class="text">Make your tap to pay program more robust with <b>rider benefits</b> & discounts</span>
```

jgravois (dismissed):
i approve as-is!

> I'm wondering if the design intention was for these "previous" and "next" links to follow the ordering of the nav links

i like this suggestion. ~lets either ask this question in Figma or in #879 when we get QA holistically.~

edit: i didn't see https://github.com/cal-itp/mobility-marketplace/pull/891#issuecomment-4292598531 until after i posted this. if @cmajel wants to chime in here, that's a-okay too.

> Open question: is the styling on the "next" link something that we're introducing as a part of this Benefits page effort? Or was that a style in Figma that was never implemented

the latter. when @Scotchester tackled this initially in #770 i _think_ he was intrepreting the design in Figma to be showing the neutral and active/onmousedown styles at the same time.

i'd like to get clarification though, because i think it'd be an improvement to differentiate the two at rest and i don't think it would take much to figure out the hover/active styles.

i'll create an issue to track that one separately.

cmajel:
@angela-tran ty for these questions!

1. **Prev/next link destinations:** That was an oversight on my part. Yes, it makes sense for this to follow the current nav flow. I've updated this in Figma to reflect this accurately
Screenshot 2026-04-22 at 1 31 06 PM

2. **Hover styles:** Ah, I had not noticed this. This was not specific to Benefits but a part of the original design which was not implemented. 
 
I don't have a strong opinion but since we have the time, I have a slight preference to implement this as designed and update the hover styles to give it distinction on the page from the more subtle buttons. I added a [demo of the different states](https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=13265-6154&t=CU2v0gz8xZ38KFHS-0) in the components section

related-right

angela-tran:
Awesome, thank you @cmajel and @jgravois! I'll update the previous/next links accordingly, including the emphasis as y'all mentioned in https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv?node-id=12365-6183&m=dev#1730255605

I'll leave the hover styles to be done in a separate PR for #892

src/data-plans.html
jgravois, line 1: the /data-plans next link should point to /route-scheduling, correct?

angela-tran: Yep that is correct. No change was needed for the /data-plans next link

jgravois: forgive me if i'm out to lunch, but isn't the /data-plans next link pointing to /gtfs-realtime right now?

https://github.com/cal-itp/mobility-marketplace/blob/da5b5a1090b70079d924b0003ffcaa99c785f78c/src/data-plans.html#L325-L328

jgravois: indeed i am. i was hung up on the bold **GTFS** text. 😂

jgravois (approved):

src/data-plans.html
angela-tran: :joy: All good!

PR #894, opened 2026-04-22
Style: related-nav-link hover and focus states

### LINKED ISSUE #892: Update the hover styles for next and previous related guide links
our [Figma designs](https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=13265-6154&m=dev) for related guide links show both the default (left) and hover (right) styles side by side.

Screenshot 2026-04-22 at 1 31 06 PM

the default styles and onclick/active/focused styles are implemented correctly, but the hover state doesn't have the correct background color.

incorrect:
Image

now that @cmajel has meticulously documented [each individual state for the component]((https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=13265-6154&m=dev)), lets update the implementation to match.

additional info in https://github.com/cal-itp/mobility-marketplace/pull/891#issuecomment-4298651635

angela-tran:
Closes #892 
(more details in the ticket)

In this branch:
**Hover**
Screenshot from 2026-04-22 14-49-44

**Focus**

image

### DIFF
--- a/src/styles/related-nav.css
+++ b/src/styles/related-nav.css
@@ -15,20 +15,22 @@
   color: var(--button-text-color);
   text-decoration: none;
 }
-.related-nav-link:focus {
-  background-color: var(--dsdl-cyan-60);
-  outline: var(--button-focus-outline);
-  outline-offset: var(--button-focus-outline-offset);
-  color: white;
+.related-nav-link:active,
+.related-nav-link:focus,
+.related-nav-link:hover {
+  color: var(--button-active-text-color);
 }
+.related-nav-link:focus,
 .related-nav-link:hover {
   background-color: var(--dsdl-cyan-60);
-  color: white;
+}
+.related-nav-link:focus {
+  outline: var(--button-focus-outline);
+  outline-offset: var(--button-focus-outline-offset);
 }
 .related-nav-link:active {
   border-color: var(--button-active-border-color);
   background-color: var(--button-active-bg-color);
-  color: var(--button-active-text-color);
 }
 .related-nav-link > .text {
   flex-grow: 1;
--- a/src/styles/related-nav.css
+++ b/src/styles/related-nav.css
@@ -16,8 +16,10 @@
   text-decoration: none;
 }
 .related-nav-link:focus {
+  background-color: var(--dsdl-cyan-60);
   outline: var(--button-focus-outline);
   outline-offset: var(--button-focus-outline-offset);
+  color: white;
 }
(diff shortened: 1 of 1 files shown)


jgravois (commented):
i think what you've got here looks great.

i'd be into consolidating the duplicate declarations and leaning a bit more on canned variables, but if you don't think 👇 is more legible, i'm not fussed.

```css
.related-nav-link {/* */}
.related-nav-link:active,
.related-nav-link:focus,
.related-nav-link:hover {
  color: var(--button-active-text-color);
}
.related-nav-link:focus,
.related-nav-link:hover {
  /* maybe even canonize as --button-related-nav-hover-bg-color? */
  background-color: var(--dsdl-cyan-60);
}
.related-nav-link:focus {
  outline: var(--button-focus-outline);
  outline-offset: var(--button-focus-outline-offset);
}
.related-nav-link:active {
  border-color: var(--button-active-border-color);
  background-color: var(--button-active-bg-color);
}
.related-nav-link > .text {/* */}
```

angela-tran:
> i think what you've got here looks great.
> 
> i'd be into consolidating the duplicate declarations and leaning a bit more on canned variables, but if you don't think 👇 is more legible, i'm not fussed.
> 

Thanks, this consolidation sounds good to me. Made the change in https://github.com/cal-itp/mobility-marketplace/pull/894/commits/36e8d6710fe2cb4aea2356e4cf12dda5697289f7 

I decided not to create a new variable for the background-color... I think it's a little questionable that we're reusing the `--button...` variables for these `related-nav-link` properties. To avoid opening a can of worms of a discussion though, I think just having the background-color use the DSDL variable inline is the cleanest way to do it. :woman_shrugging:

jgravois (approved):
it would make sense to me to change the base branch of this PR to `main`, but I approve regardless. 😇

angela-tran:
Yeah... I agree we've had a few PRs that are in a bit of a gray area in terms of "belonging" to the Benefits page or not. I feel uneasy whenever we have to rebase `feat/riders-benefits` onto `main` since I just imagine that not being a workable model if we were working on a feature with lots of developers doing parallel work. But maybe I'm just trying to generalize when I shouldn't. And it may be unavoidable in any case that there are changes needed in `main`.

Anyways, definitely a thing to get more aligned on in the future! Thanks for mentioning it

I'll go ahead and merge into `feat/rider-benefits` to avoid the rebase

PR #895, opened 2026-04-23
Chore: replace "Tap2Pay" with "Tap to pay"

angela-tran:
Closes #893 

A simple search and replace

PR #905, opened 2026-04-27
Style: allow more room for pull-quote attribution

### LINKED ISSUE #899: Adjust pull quote attribution sizing
There is a difference in the pull quote attribution sizing between Figma and production that results in more line breaks. 

This work would to make an adjustment at the component level to the attribution sizing. After doing some exploring, I believe swapping `col-lg-4` for `col-lg-5` for that div in `figcaption.row` gets us pretty close.

### Production

Image

### Figma

Image

angela-tran:
Closes #899 

This PR applies the suggested change to go from `col-lg-4` to `col-lg-5` for the div holding the quote attribution. It works well!

It makes the width match Figma better for not only `/rider-benefits`, but also `/gtfs-realtime` and `/tap-to-pay`.

<details><summary> Screenshots</summary>

`/rider-benefits`

image

`/gtfs-realtime` and `/tap-to-pay`

| Live | This branch | Figma |
| --- | --- | --- |
| Screenshot from 2026-04-27 13-08-44 | Screenshot from 2026-04-27 13-08-49 | Screenshot from 2026-04-27 13-09-16 |
| Screenshot from 2026-04-27 13-09-35 | Screenshot from 2026-04-27 13-09-37 | Screenshot from 2026-04-27 13-09-30 |

</details>

## Some notes on the target branch

Similar to changes like the ones in #894, #895, #886, this work wasn't exclusively for implementing the Benefits page, but rather was more to fix a discrepancy between what's in prod vs. in Figma that we noticed because we're implementing the Benefits page. However, another way to look at it is the Benefits page implementation could not be complete without fixing those discrepancies.

This makes it ambiguous whether we should target `main` or `feat/rider-benefits`, the "staging" branch we've been working off. I think which one feels more "correct" essentially depends on how you feel about 
- rebasing the staging branch onto `main`
- whether reviewers should be able to see all changes in the staging branch preview site
- whether those discrepancy fixes should be deployed to the live site before the new page

The discussion in https://github.com/cal-itp/mobility-marketplace/pull/894#pullrequestreview-4158175939 adds more context.

It would be great to get consensus about our team philosophy on this sort of branching situation so that in the future we can move more confidently and don't have to try to figure this out on-the-fly. I recall this coming up in the past with websites work, and we've just never really talked through this.

I have this branch targeting `feat/rider-benefits` so that @leelas-pm and @cmajel can review it on the staging branch preview site since this was feedback on the Benefits page.

### DIFF
--- a/src/_includes/pull-quote.html
+++ b/src/_includes/pull-quote.html
@@ -28,7 +28,7 @@
     </blockquote>
     {% if source_name %}
       <figcaption class="row">
-        <div class="col-lg-4 offset-lg-6">
+        <div class="col-lg-5 offset-lg-6">
           <div class="fw-bold">{{ source_name }}</div>
           {% if source_title %}
             <div>{{ source_title }}</div>


jgravois (approved):
lgtm!

> It would be great to get consensus about our team philosophy on this sort of branching situation

from where i stand there's no downside to landing independent improvements like this directly to `main` and then rebasing a long-running branch afterward.

that said, i don't have a problem with a change like this coming along for the ride in #879 after being reviewed/approved in a PR of its own.

i prefer the former option, but i'd call the difference "[small beer](https://en.wikipedia.org/wiki/Small_beer#Literature)". the same commit is going to be accessible in the history on `main` eventually regardless.

PR #909, opened 2026-04-29
Fix: rewire Google Analytics

### LINKED ISSUE #908: bug: rewire google analytics
analytics events stopped being logged back in Feb 2026. we're still injecting the correct key and making a call to the tag manager endpoint, so something else must be amiss.

Image

we need to make sure that the approach to tag injection below is sound.
https://github.com/cal-itp/mobility-marketplace/blob/d11a3ef71ae913c30bf377ded555f1e0ba532266/src/_layouts/main.html#L25-L40

angela-tran:
Closes #908 

See https://github.com/cal-itp/mobility-marketplace/issues/908#issuecomment-4346880998 for notes

### DIFF
--- a/src/_layouts/main.html
+++ b/src/_layouts/main.html
@@ -22,22 +22,17 @@
     <link rel="stylesheet" href="/styles/main.css">
     {% if stylesheet %}<link rel="stylesheet" href="/styles/{{ stylesheet }}">{% endif %}
 
-    <!--
-      Google Tag Manager https://support.google.com/tagmanager/answer/14847097?hl=en&ref_topic=15191151&sjid=3924574870016540527-NC
-    -->
+    <!-- Google tag (gtag.js) -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id={{ site.google_analytics_id }}"></script>
     <script>
-      (function (w, d, s, l, i) {
-        w[l] = w[l] || [];
-        w[l].push({ 'gtm.start': new Date().getTime(), event: 'gtm.js' });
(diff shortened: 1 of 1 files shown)


jgravois (commented):
awesome!

if the genuine doc includes mention of a `<no script>` fallback that would be good to add too, but I don't consider that a blocker.

angela-tran:
> awesome!
> 
> if the genuine doc includes mention of a `<no script>` fallback that would be good to add too, but I don't consider that a blocker.

It doesn't mention it, but I think it's reasonable to add it too :+1:

src/_layouts/main.html
```diff
@@ -22,24 +22,30 @@
     <link rel="stylesheet" href="/styles/main.css">
     {% if stylesheet %}<link rel="stylesheet" href="/styles/{{ stylesheet }}">{% endif %}
 
-    <!--
-      Google Tag Manager https://support.google.com/tagmanager/answer/14847097?hl=en&ref_topic=15191151&sjid=3924574870016540527-NC
-    -->
+    <!-- Google tag (gtag.js) -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id={{ site.google_analytics_id }}"></script>
```
angela-tran, line 26: Still not sure why the previous code wouldn't work, but I do note the difference of

`https://www.googletagmanager.com/gtm.js?id=`

vs.

`https://www.googletagmanager.com/gtag/js?id=`

jgravois (approved):
lets give it a shot!

PR #912, opened 2026-04-30
chore: re-enable prettier pre-commit hook

### LINKED ISSUE #898: chore: reintroduce prettier pre-commit hook
while refactoring calitp.org to build w/ 11ty we figured out how to get VS Code and pre-commit to agree on how to autoformat files. lets implement those changes here too. 

## Acceptance criteria
- [ ] prettier is added to the package.json file
- [ ] prettier autoformat .html, .md, .json, etc. in VS Code
- [ ] @shopify/prettier-plugin-liquid autoformats .liquid files in VS Code
- [ ] new npm scripts are added to bulk check/fix files
- [ ] an additional pre-commit check verifies formatting (using the same configuration)
- [ ] VS Code is configured such that prettier autoformats html, markdown, json, etc.
- [ ] files with liquid content are given the `.liquid` extension and references are updated accordingly
- [ ] no visual regressions are introduced in the site

prior art:
* https://github.com/cal-itp/calitp.org/pull/596
* https://github.com/cal-itp/calitp.org/pull/600

### LINKED ISSUE #911: chore: re-enable prettier pre-commit hook
right now VS Code uses prettier and a shopify liquid extension to autoformat on save.

after the initial 11ty refactor here we figured out a viable technique in calitp.org (ref: https://github.com/cal-itp/calitp.org/pull/600) to get pre-commit to enforce the same formatting on commit.

we should backport that improvement here.

jgravois:
closes #898

* the first commit in this PR fixes a few instances of invalid html that the prettier cli caught out
* the second updates the prettier and VS Code configs for pre-commit compatibility
* the third declares the bulk of the static files for the site as `.liquid` explicitly and runs prettier autoformatting in bulk (this didn't result in many changes)
* the last enables the prettier pre-commit hook.

i learned something _really_ interesting implementing this fix. it turns out that pre-commit looks for a `.node-version` and uses that version when present to lay down its own runtime.

this is great because it ensures parity with what happens elsewhere in the dev container and on Netlify. the only catch is that its not able to interpret a bare value like '24' and look up the latest and greatest patch release on the fly.

imo, its not a bad tradeoff to be more explicit in our own .node-version file. ~in the future we can/should make a similar tweak in calitp.org instead of declaring the version of node we'd like to use in the [mise.toml](https://github.com/cal-itp/calitp.org/blob/2bb6f89bbde96f2731231cf8f3017ffb1c10187d/mise.toml#L1-L2) file~. 
(☝️ was completed in https://github.com/cal-itp/calitp.org/pull/625)

## Steps to test

we'll need to confirm that this PR doesn't introduce any visual regressions in the live site.


.vscode/settings.json
```diff
@@ -7,17 +7,12 @@
   "files.insertFinalNewline": true,
   "files.trimFinalNewlines": true,
   "files.trimTrailingWhitespace": true,
-  "files.associations": {
-    "*.html": "liquid"
-  },
-  "[javascript]": {
-    "editor.tabSize": 2
+  // the declarations below take precedence over file specific formatters specified in *user* settings
```
jgravois, line 10: despite the scary red text here, i confirmed that VS Code [is indeed able to ignore comments in .json configuration files](https://www.reddit.com/r/vscode/comments/16rfw5p/why_does_settingsjson_support_comments/). 😇

src/_includes/seo.liquid
jgravois, line 1: this refactor taught me that shopify's liquid extension doesn't inject the trailing slashes that @Scotchester despises into void html tags.

```html
.liquid 😇 
<meta ...>

.html 🤮 
<meta .../>
```

angela-tran:
I plan to review this / test it out tomorrow!

Scotchester:
_[I canceled the Copilot review that I accidentally requested.]_

Scotchester (dismissed):
I compared the production site and the PR preview site side-by-side and everything looks unchanged, so that part's good.

I have two concerns, though:

#### Prettier not running in VS Code

When I try to format, either on save or manually, Prettier is throwing this error:

```
["INFO" - 7:26:16 PM] Using config file at /usr/src/mobimart/.prettierrc.js
["INFO" - 7:26:16 PM] EditorConfig support is enabled, checking for .editorconfig files
["ERROR" - 7:26:16 PM] Invalid prettier configuration file detected. See log for details.
["ERROR" - 7:26:16 PM] Cannot find module '@shopify/prettier-plugin-liquid'
Require stack:
- /usr/src/mobimart/.prettierrc.js
Error: Cannot find module '@shopify/prettier-plugin-liquid'
```

As mentioned in Slack, I have just completely started my MobiMart environment from scratch deleted all Docker containers and images for good measure). I can also see that `@shopify/prettier-plugin-liquid` **is** present in `node_modules`. Not sure what to make of it.

Good news, though: pre-commit _does_ correctly reformat the file if I try to commit it.

#### Commit history

> - the third declares the bulk of the static files for the site as .liquid explicitly
> - the fourth runs prettier autoformatting in bulk (this didn't result in many changes)

I'm requesting changes because it looks to me like the bulk of the formatting actually happens in the third commit (alongside the renaming of `.html` → `.liquid`, and the fourth commit actually reverts some of those changes in `contact.liquid` (correctly, given the end result we want).

I think probably the easiest thing to do would be to squash commits 3 and 4 together and add the resulting commit to `.git-blame-ignore-revs`. (I don't think there's much point in stepping through this extension swap with no changes when blaming, so let's just include it together with the formatting.)

jgravois:
> When I try to format, either on save or manually, Prettier is throwing this error:

i cannot repro that. are you still pulling `python:3.12-bookworm`? regardless, lets take a look at your setup together next week.

> the easiest thing to do would be to squash commits 3 and 4 together and add the resulting commit to .git-blame-ignore-revs

i think this is a great suggestion. i'll wait until #915 and #917 land before i rewrite history though.

Scotchester:
Yeah, still using bookworm. Pairing Tuesday would be great.

jgravois:
> i'll wait until https://github.com/cal-itp/mobility-marketplace/pull/915 and https://github.com/cal-itp/mobility-marketplace/pull/917 land before i rewrite history

done!

.vscode/settings.json
```diff
@@ -7,17 +7,12 @@
...
+  "[html]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "[json]": {
-    "editor.tabSize": 2
-  },
-  "[liquid]": {
-    "editor.defaultFormatter": "Shopify.theme-check-vscode"
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
angela-tran, line 21: I suggest we go ahead and include all the languages that we [list in our pre-commit hook](https://github.com/cal-itp/mobility-marketplace/pull/912/changes#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R35) for consistency with it and our other projects - like 

https://github.com/cal-itp/calitp.org/blob/caf7fdb4a4ed83105d1beb99a44518b9944f5d98/.vscode/settings.json#L13-L15 

(as a follow-up, we should fix that calitp.org `settings.json` file to separate out the languages like in https://github.com/compilerla/compiler.la/blob/1e56a3cda084ce656246d21dc367fa7bbabdb00f/.vscode/settings.json#L10-L37)

angela-tran:
```suggestion
  // the declarations below take precedence over file specific formatters specified in *user* settings
  "[css]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[html]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[javascript]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[json]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[jsonc]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[liquid]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[markdown]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[yaml]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  }
```

.vscode/settings.json
```diff
@@ -7,17 +7,12 @@
...
-    "editor.tabSize": 2
+  // the declarations below take precedence over file specific formatters specified in *user* settings
+  "[html]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "[json]": {
-    "editor.tabSize": 2
-  },
-  "[liquid]": {
-    "editor.defaultFormatter": "Shopify.theme-check-vscode"
```
angela-tran, line 20: Makes sense that we're removing this from being used as the formatter but still keeping it as an extension (in `.devcontainer.json`) to get the other features from it, e.g. syntax highlighting, etc. 

(https://github.com/cal-itp/calitp.org/pull/596#issuecomment-4194626146 helped remind me of this.)

As a separate follow-up, it'd be good to align which extensions we're using across our static site projects. i.e. resolve the drift between the TOML extensions and Liquid extensions.

angela-tran (changes requested):
This looks great to me @jgravois 

I was able to see that VS Code reformats file contents as expected and that the pre-commit hook prevents me from committing badly formatted code.

 There was just a nitpick I had on the `editor.defaultFormatter` overrides, but aside from that, this looks good to go!

angela-tran:
(Also, oof on that invalid HTML, thanks for fixing those!)

.vscode/settings.json
jgravois: good call! f36f3a6

angela-tran (approved):

.vscode/settings.json
jgravois: i agree! done in #922

PR #915, opened 2026-05-19
Refactor: split /contact and /about pages and update nav

### LINKED ISSUE #914: seperate /contact and /about into two distinct pages
> * Creates distinct [About](https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/branch/BbQZeMUm6Lc1BVlSQgHfel/Mobility-Marketplace?node-id=13403-6504&t=KEUgMKZNiEaiuohE-0) and [Contact](https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/branch/BbQZeMUm6Lc1BVlSQgHfel/Mobility-Marketplace?node-id=13403-6452&t=KEUgMKZNiEaiuohE-0) pages (Currently this is all on the Contact page)

> * Updates desktop and mobile navigation to add new guides and a subnav for About / Contact us
>   * I added a [dev notes page with annotations](https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/branch/BbQZeMUm6Lc1BVlSQgHfel/Mobility-Marketplace?node-id=13403-16162&p=f&t=KEUgMKZNiEaiuohE-0) to describe the changes here. They include
>     * Adding a subnav for About/Contact us
>     * On mobile:
>       * Adding a close icon to the mobile nav to make it clearer the button does something different when the nav is opened
>       * Styling all links (subnav and guides) the same, to make it clearer they are all links
>       * Adjusting spacing slightly for easier taps and to distinguish the nav categories
>       * Adding borders to the top and bottom of the opened nav to distinguish it from page content

as originally articulated by @cmajel in https://github.com/cal-itp/mobility-marketplace/issues/795#issuecomment-4329954743

jgravois:
closes #914

view it live: 
* https://deploy-preview-915--cal-itp-mobility-marketplace.netlify.app/about/
* https://deploy-preview-915--cal-itp-mobility-marketplace.netlify.app/contact/

### DIFF
--- a/src/_includes/header/desktop.html
+++ b/src/_includes/header/desktop.html
@@ -1,4 +1,14 @@
-<header class="d-none d-lg-flex flex-column header mx-auto">
+<header class="d-none d-lg-flex flex-column-reverse header mx-auto">
+  <div class="d-flex justify-content-between align-items-center gap-3">
+    {% include 'site-logo.html' %}
+    <div class="d-flex flex-column align-items-end gap-3">
+      <nav class="global-nav" aria-label="Primary">
+        <ul class="d-flex justify-space-between">
+          {% include 'link-repeater.html', menu_items: site.navigation, list_class: 'mb-0' %}
+        </ul>
+      </nav>
+    </div>
+  </div>
   <div class="d-flex flex-column align-items-end gap-3">
     <nav class="d-flex gap-3">
       <a
@@ -20,15 +30,5 @@
         Contact
       </a>
     </nav>
-  </div>
-  <div class="d-flex justify-content-between align-items-center gap-3">
-    {% include 'site-logo.html' %}
-    <div class="d-flex flex-column align-items-end gap-3">
-      <nav class="global-nav" aria-label="Primary">
-        <ul class="d-flex justify-space-between">
-          {% include 'link-repeater.html', menu_items: site.navigation, list_class: 'mb-0' %}
-        </ul>
-      </nav>
-    </div>
   </div>
 </header>
--- a/src/_includes/header/desktop.html
+++ b/src/_includes/header/desktop.html
@@ -1,17 +1,30 @@
 <header class="d-none d-lg-flex align-items-center justify-content-between header mx-auto">
   {% include 'site-logo.html' %}
-  <nav class="global-nav" aria-label="Primary">
-    <ul class="d-flex justify-space-between">
-      {% include 'link-repeater.html', menu_items: site.navigation list_class: 'mb-0' %}
-    </ul>
-  </nav>
-  <a
-    class="contact font-weight-bold"
-    href="/contact"
-    {% if page.url == '/contact.html' %}
-      aria-current="page"
-    {% endif %}
-  >
-    Contact
-  </a>
+  <div class="d-flex flex-column align-items-end gap-3">
+    <nav class="d-flex gap-3">
+      <a
+        class="contact"
+        href="/about/"
+        {% if page.url == '/about/' %}
+          aria-current="page"
+        {% endif %}
+      >
+        About
+      </a>
+      <a
+        class="contact"
+        href="/contact/"
+        {% if page.url == '/contact/' %}
+          aria-current="page"
+        {% endif %}
+      >
+        Contact
+      </a>
+    </nav>
+    <nav class="global-nav" aria-label="Primary">
+      <ul class="d-flex justify-space-between">
+        {% include 'link-repeater.html', menu_items: site.navigation, list_class: 'mb-0' %}
+      </ul>
+    </nav>
+  </div>
 </header>
--- a/src/_includes/header/desktop.html
+++ b/src/_includes/header/desktop.html
@@ -1,14 +1,4 @@
-<header class="d-none d-lg-flex flex-column-reverse header mx-auto">
-  <div class="d-flex justify-content-between align-items-center gap-3">
-    {% include 'site-logo.html' %}
-    <div class="d-flex flex-column align-items-end gap-3">
-      <nav class="global-nav" aria-label="Primary">
-        <ul class="d-flex justify-space-between">
-          {% include 'link-repeater.html', menu_items: site.navigation, list_class: 'mb-0' %}
-        </ul>
-      </nav>
-    </div>
-  </div>
+<header class="d-none d-lg-flex flex-column header mx-auto">
   <div class="d-flex flex-column align-items-end gap-3">
     <nav class="d-flex gap-3">
       <a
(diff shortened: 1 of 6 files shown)


src/_includes/header/desktop.html
```diff
@@ -1,17 +1,30 @@
 <header class="d-none d-lg-flex align-items-center justify-content-between header mx-auto">
   {% include 'site-logo.html' %}
-  <nav class="global-nav" aria-label="Primary">
-    <ul class="d-flex justify-space-between">
-      {% include 'link-repeater.html', menu_items: site.navigation list_class: 'mb-0' %}
-    </ul>
-  </nav>
-  <a
-    class="contact font-weight-bold"
-    href="/contact"
```
jgravois, line 10: trivial, but adding a trailing slash here sidesteps an unnecessary redirect

src/_includes/header/desktop.html
```diff
@@ -1,17 +1,30 @@
...
+        {% if page.url == '/about/' %}
+          aria-current="page"
+        {% endif %}
+      >
+        About
+      </a>
+      <a
+        class="contact"
+        href="/contact/"
+        {% if page.url == '/contact/' %}
```
jgravois, line 16: changing this from `/contact.html` fixes a bug we introduced when migrating to 11ty that prevented the current page from being advertised as such.

Image

src/styles/header.css
```diff
@@ -98,3 +97,11 @@
 .navbar-nav > li {
   margin-bottom: 0;
 }
+
+.navbar-toggler[aria-expanded="true"] .navbar-toggler-icon {
+  background-image: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='-4 -4 24 24' fill='rgba(33, 37, 41, 0.75)'><path d='M.293.293a1 1 0 0 1 1.414 0L8 6.586 14.293.293a1 1 0 1 1 1.414 1.414L9.414 8l6.293 6.293a1 1 0 0 1-1.414 1.414L8 9.414l-6.293 6.293a1 1 0 0 1-1.414-1.414L6.586 8 .293 1.707a1 1 0 0 1 0-1.414'/></svg>");
+}
```
jgravois, line 107: i hacked at this <svg> myself. @cmajel if you'd like to get more finegrained control over either the <kbd>X</kbd> or the 🍔, it'd definitely be straightforward to implement it here in the CSS.

src/styles/header.css
```diff
@@ -98,3 +97,11 @@
...
+
+.header:has(.navbar-toggler[aria-expanded="true"]) {
+  border-bottom: 1px solid var(--dsdl-gray-30);
+}
```
jgravois, line 111: this is [somewhat modern](https://caniuse.com/?search=css+has) CSS, but AFAIK the fallback is a benign no-op.

src/about.html
```diff
@@ -0,0 +1,36 @@
+---
+layout: main
+
+title: About
+stylesheet: guide.css
+description: TK (maybe)
```
jgravois: whoops, forgot about this one.

@cmajel we either need to come up with a custom description for this page to display in shareable thumbnail previews or nix it entirely in favor of the generic site description.

https://github.com/cal-itp/mobility-marketplace/blob/11bca90da7b8784deb8b54156af396ac8a78cb40/src/_data/site.js#L3-L4 

i'm fine with either option personally.

cmajel: Ah yes, I've added a suggestion to the Figma file and reposting here (feedback welcome). We should aim to have unique page descriptions for better search experiences.

>Learn why we created the California Mobility Marketplace and how we support transit agencies and the riders they serve.

cmajel:
Thanks @jgravois!

**nav borders**
I'm ok dropping that border if it's a huge amount of work to get it in there. 

**logo alignment**
On the current version the keyboard experience is what we want and expect, so I get how you landed here. 

I would like to see if we can get closer to the design without losing the tab ordering. With the current alignment on desktop, the logo draws you toward all the nav items (sub and main), when the intention was to align it to give more attention to the main nav. 

If it's overly difficult to refactor I can look into other options with this structure.

Scotchester:
This was already the situation before this PR, but while we're messing around in the mobile menu, this whole column probably shouldn't be a live target area for the home link.

Dev tools highlighting the home link area on a mobile-width screen

As for the border… an absolutely positioned `::before` pseudoelement would be the quickiest, hackiest way I can think of to address it.

src/about.html
jgravois: implemented in 5900216. thx!

jgravois:
> I would like to see if we can get closer to the design without losing the tab ordering

thanks for the nudge. this was actually trivial. 🤦‍♂️  5900216 (~and 662bf12~)

> an absolutely positioned ::before pseudoelement would be the quickiest, hackiest way I can think of to address it.

is 725283f what you had in mind @Scotchester?

(i had to futz with this for longer than i care to admit, but constructive criticism is still welcome)

> this whole column probably shouldn't be a live target area for the home link.

great catch. fixed in fd1fd60

Scotchester:
> > an absolutely positioned ::before pseudoelement would be the quickiest, hackiest way I can think of to address it.
> 
> is 725283f what you had in mind @Scotchester?
> 
> (i had to futz with this for longer than i care to admit, but constructive criticism is still welcome)

Yeah, basically exactly that! I wouldn't mind applying it to `.header` to avoid creating a new element, and you could simplify these two properties:

```css
width: 100%;
left: 0;
```

But it looks as intended :+1:

cmajel:
Nice, looks much more like what we're going for. 

Two small adjustments:
* I do see this adjusted the tab order, and on keyboard you go through the main nav before the subnav. Is it possible to adjust this so keyboard users go through the subnav, then the main nav? My thinking is to match the visual hierarchy for keyboard users, but open to thoughts on this. 

* In Figma there is a wider gap (~40px / `spacing-5`) between the nav and inner page content (guides, about, contact, docs). Can we make that adjustment either in this PR or in a follow up?

### Current:
Screenshot 2026-05-26 at 11 41 32 AM

Screenshot 2026-05-26 at 11 41 23 AM


### Adjusted:
Screenshot 2026-05-26 at 11 41 07 AM
Screenshot 2026-05-26 at 11 41 15 AM

Scotchester:
> I do see this adjusted the tab order, and on keyboard you go through the main nav before the subnav. Is it possible to adjust this so keyboard users go through the subnav, then the main nav? My thinking is to match the visual hierarchy for keyboard users, but open to thoughts on this.

That's my understanding of the best practice, too.

jgravois:
> Is it possible to adjust this so keyboard users go through the subnav, then the main nav?

sorry. i misunderstood completely last week. i thought you were saying you wanted to _preserve_ users tabbing into the guides before the sub-nav. fixed in e1b1072

> In Figma there is a wider gap (~40px / spacing-5) between the nav and inner page content

great catch. fixed in e1b1072

cmajel (approved):
Ah sorry for the misunderstanding re tab order. 

Changes look great - approved! 🎉

PR #924, opened 2026-06-04
chore: constrain devcontainer user permissions

### LINKED ISSUE #920: chore: update devcontainer setup to use non-root user
lets emulate the [benefits devcontainer](https://github.com/cal-itp/benefits/blob/baea4e0018802b53a04254a0130bf965bca5c5ca/.devcontainer/Dockerfile#L2) configuration and ensure that root privileges are inaccessible after the container is built.

since we're not pulling `docker-python-web` here, we'll need to do the same legwork as is done here 👇 

https://github.com/compilerla/compiler.la/blob/main/.devcontainer/Dockerfile#L7-L8

jgravois:
closes #920

analogous to https://github.com/cal-itp/calitp.org/pull/652


.devcontainer/postAttach.sh
```diff
@@ -1,13 +1,15 @@
 #!/usr/bin/env bash
 set -eux
 
-git config --global --add safe.directory /usr/src/mobimart
+git config --global --add safe.directory /home/mobimart/site
```
thekaveman: I guess since we aren't installing a Python package that needs git Metadata, this is fine to do in `postAttach` vs. in the Dockerfile? I saw the same thing in https://github.com/cal-itp/calitp.org/pull/652 and seems like it worked out OK there.

.devcontainer/Dockerfile
```diff
@@ -18,3 +23,5 @@ RUN apt-get update
 COPY .devcontainer/requirements.txt .devcontainer/requirements.txt
 RUN python -m pip install --upgrade pip
 RUN pip install --no-cache-dir -r .devcontainer/requirements.txt
+
+USER $USER
```
angela-tran, line 28: _(no change needed, just jotting down some notes)_

I've been wondering (for this PR and for cal-itp/calitp.org#652) about this line from the compiler.la repo and whether we need it or not:

https://github.com/compilerla/compiler.la/blob/e717571f5b8263019926430d6f68d73a189add7a/.devcontainer/Dockerfile#L41

```diff
USER $USER
+ ENV PATH "$PATH:/home/$USER/.local/bin"
```

I would guess it [was added](https://github.com/compilerla/compiler.la/commit/3ba1b24f6bfd0f659bdec90c5d891ea641ab8453#diff-13bd9d7a30bf46656bc81f1ad5b908a627f9247be3f7d76df862b0578b534fc6R17-R23) because that's where a user install of [Python](https://docs.python.org/3/library/sysconfig.html#sysconfig-user-scheme)/[pip](https://pip.pypa.io/en/latest/user_guide/#user-installs) would go.

However, when we look at the Python path in the devcontainer for any of these 3 repos, I see that it's using the system-install path.
```
python
> import sys
> print(sys.path)
```
outputs
```
['', '/usr/local/lib/python312.zip', '/usr/local/lib/python3.12', '/usr/local/lib/python3.12/lib-dynload', '/usr/local/lib/python3.12/site-packages']
```

Also, when I run `pip list -v`, that shows that Python packages are being installed to the system install as well.

So I don't think we need that line here or in calitp.org or compiler.la.

angela-tran: > I see that it's using the system-install path

> Python packages are being installed to the system install as well.

I think this makes sense because the base image already has Python installed, and we're running `pip install` before switching to the non-root user.

.devcontainer/postAttach.sh
jgravois: exactly. it works fine as-is.

now that you mention it though, beyond being more consistent, moving this line into the Dockerfile would also allow us to us to reference `$USER` instead.

i don't love all the duplication in this PR (and the equivalent code in compiler.la and calitp.org) but i played around with defining a USER variable in the compose.yml file and devcontainer.json in the name of centralizing and couldn't find an approach that'd work everywhere.

ref: https://code.visualstudio.com/remote/advancedcontainers/environment-variables

.devcontainer/Dockerfile
jgravois: i agree. when i saw it in compiler.la i did enough diligence to convince myself that it wasn't needed here but based on your write-up i'm convinced its superfluous there now too. ❤️

.devcontainer/postAttach.sh
jgravois: > moving this line into the Dockerfile would also allow us to us to reference $USER instead.

done in 0cbb3c7

.devcontainer/Dockerfile
thekaveman: I think if we were working on Python projects (like Benefits) we'd want to `pip install --user` e.g.

https://github.com/cal-itp/benefits/blob/c004b61a3462f4c4552af188bb59a0f2778d5cb1/appcontainer/Dockerfile#L120

And

https://github.com/cal-itp/benefits/blob/c004b61a3462f4c4552af188bb59a0f2778d5cb1/.devcontainer/Dockerfile#L40

But agreed this is fine for static sites (if it works, I haven't verified :sweat_smile:)

.devcontainer/Dockerfile
```diff
@@ -1,20 +1,30 @@
 FROM python:3.12
 
-WORKDIR /usr/src/mobimart
+ENV USER=mobimart
+ENV HOME=/home/$USER
+
+RUN useradd --create-home --shell /bin/bash $USER && \
+    chown -R $USER $HOME
+
+WORKDIR $HOME/site
```
Scotchester, line 9: This made me realize that we use a shorter WORKDIR on benefits, [inherited from docker-python-web](https://github.com/cal-itp/docker-python-web/blob/ba941dea0e026ef92c23a443f656f9a214790a38/appcontainer/Dockerfile#L96):

```Dockerfile
WORKDIR /$USER/app
```

Do we want to standardize on one or the other?

Scotchester (commented):
This seems like a local issue, because nothing in this PR seems like it would cause it, but for some reason the `postAttachCommand` is not running for me 🤔

Scotchester:
> This seems like a local issue, because nothing in this PR seems like it would cause it, but for some reason the `postAttachCommand` is not running for me 🤔

On the other hand, it still works fine on `main`…

.devcontainer/Dockerfile
angela-tran: I don't have a strong preference for standardizing or not, but just for context, the shorter WORKDIR was motivated by needing to reserve the `/home` directory for the Azure App Service. Azure gives the app service a persistent storage space that is seen by the app container as part of its file system., and the whole reason we needed that was to hold our SQLite database. Mounting a network file share to be used by the app container was not stable for SQLite.

See https://github.com/cal-itp/docker-python-web/commit/5f949caefd81a1afab9e2b0b2875d6005e400be0 and its PR https://github.com/cal-itp/docker-python-web/pull/42

jgravois:
> On the other hand, it still works fine on `main`…

perhaps there's a line failing and the exception is being swallowed? 🤔

.devcontainer/Dockerfile
angela-tran: For our static sites, maybe it's better to just leave it as-is with `$HOME` since that seems like a reasonable (and intuitive?) default and there's no motivating reason to do something different. :woman_shrugging:

Scotchester: Cool with me. Thanks for the background, Angela!

angela-tran:
I was able to do a clean rebuild of the devcontainer from this branch. I'm curious what error you're seeing @Scotchester

Scotchester:
No error at all; it's just not running the `postAttachCommand` (which I noticed because I went to run the task to launch the site and it failed due to missing `npm`).

lalver1:
I'm seeing the same behavior with `postAttach.sh` not running (maybe failing silently?) because the dev container starts, but like Scott mentioned, it's missing `npm` so the `11ty: Serve` task does not work. I noticed that if I move `git config --global --add safe.directory /home/mobimart/site` back to `postAttach.sh` I can avoid this problem but I'm not sure yet why this works. I'm looking at this a bit more to make sure I understand what may be going on, but my guess is something related to who the active user is when `git config --global --add safe.directory /home/mobimart/site` runs in the Dockerfile (before we switch to `$USER`) compared to running the command as `$USER` in `postAttach.sh`, maybe 🤔 .

jgravois:
ah, interesting find @lalver1.

my instinct is to just leave the command in the postAttach script until/unless we encounter a problem like https://github.com/cal-itp/benefits/pull/3630 but if anyone else has a better idea, i'm all ears. 👂

lalver1:
Yep, we could do that @jgravois, leave it in `postAttach.sh`. But if we want to keep it in the Dockerfile, for example to take advantage of referencing `$USER`, it looks like moving `RUN git config --global --add safe.directory $HOME` under `USER $USER` will solve the issue, i.e., configuring `git` after we have switched to the `mobimart` user in the Dockerfile.

jgravois:
@lalver1 i agree thats a superior option 😇 

@Scotchester lemme know whether you have better luck with 4a59bcc

Scotchester (approved):
Works for me now!

Issue #906, opened 2026-04-28
Investigate numbered headings being repeated on screenreaders


cmajel:
While testing a Voiceover read through of #879, I noticed that the numbered headings in guides are read twice by the screenreader.

This occurs for the numbered H3s after the pull quote (the numbered list above the pull quote is not affected). I confirmed this happens across guides and on production. Video demo below.

This doesn't seem like the right behavior, but I'm not sure why it is happening. Some comparable patterns are the [VA's process list ](https://design.va.gov/components/process-list)and the [step list from CMS ](https://design.cms.gov/v/5.0.2/storybook/iframe.html?id=patterns-step-list--step-list-example&viewMode=docs&globals=theme:core)

https://github.com/user-attachments/assets/494fc7d1-14d9-4f8d-9da0-2a19479e7323

jgravois:
it took me a little bit to figure out how you were [navigating in Voiceover](https://support.apple.com/guide/voiceover/read-text-vo2706/mac). very cool! 😅 

the gist of the problem is that its the `::before` pseudoelement that's giving us grief.

adding a fallback like this 👇 would cause Voiceover to skip reading the number entirely and only read the `<h3>` text once.

```css
.numbered::before {
  /* https://caniuse.com/wf-alt-text-generated-content */
  content: attr(data-number) / ""; 
  /* ... */
}
```

i could live with that, but i think its worth spending an hour or two on a smarter fix and try shuffling the numbers into a `<span>` instead. holler if you disagree.

ref: https://whitep4nth3r.com/blog/hide-text-in-css-pseudo-elements-from-screen-readers/

jgravois:
turns out headings in particular are known to cause this misbehavior and the Chrome + Voiceover combo is the worst offender.

https://lab.dotjay.com/tests/screen-readers/voiceover-text-breaks-workarounds/

i was able to verify ☝️. ie: nested/duplicate content is announced repetitively whether or not the additional child is a `::before` or `<span>`.

adding a `role` to the entire `<h3>` is enough to flatten things out in Safari, but not Chrome. 😞

```html
<h3 role="text">
  <span class="numbered-badge numbered-badge_purple">1</span>
  Choose a data plan
</h3>
```

see also: https://www.youtube.com/watch?v=Nkcnm7xZZP4

## our options, as i see them:

1. skip announcing the numbers in the numbered headings altogether
1. add `role="text"` and expect Safari to announce the entire string as desired and Chrome to repeat individual children
[...]

PR #929, opened 2026-06-09
fix: smooth out VoiceOver reading of numbered headings

jgravois:
the purpose of this PR is to test another POC related to #906

### Option 1: add a [wrapping `<div>`](https://deploy-preview-929--cal-itp-mobility-marketplace.netlify.app/demo/#poc)

```html
<div data-number="1">
  <h3>Choose a data plan</h3>
</div>
```
with this option, VoiceOver announces each number immediately before reading each heading in both browsers.

since the number is divorced from the actual heading, it is also omitted from the VoiceOver rotor.

https://github.com/user-attachments/assets/5a8654e6-0fa4-407a-8d43-b91f4283b30a

Screenshot 2026-06-11 at 2 43 15 PM

### Option 2: `<h3 role="text">`

the other option we discussed previously in https://github.com/cal-itp/mobility-marketplace/issues/906#issuecomment-4627026828, 👇 would smooth out the way the headings are read in VoiceOver + Safari but not VoiceOver + Chrome.

```html
<h3 role="text">
  <span class="numbered-badge">1.</span>
  Choose a data plan
</h3>
```

admittedly neither of the solutions i've come up with are a silver-bullet. i lean slightly toward option 1, but i'm going to tag @Scotchester here too in case he's got a trick up his sleeve or is available to pair on it.

### DIFF
--- a/src/demo.liquid
+++ b/src/demo.liquid
@@ -148,6 +148,16 @@ for our small team. We’re able to prioritize GTFS-RT first and foremost before
     </div>
   </div>
 
+  <div class="row">
+    <h2 id="poc">Numbered headings (w/ wrapping <code>&lt;div&gt;</code>)</h2>
+    <div class="offset-lg-2 col-lg-7 col-12">
+      <div class="vo-friendly" data-number="1"><h3>Choose a data plan</h3></div>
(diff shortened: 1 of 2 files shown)


jgravois:
closing in favor of #934

PR #934, opened 2026-06-16
fix: use an ordered list to smooth out VoiceOver reading of numbered headings

### LINKED ISSUE #906: Investigate numbered headings being repeated on screenreaders

jgravois:
closes #906

### The solution: [wrap all numbered headings in an `<ol>`](https://deploy-preview-934--cal-itp-mobility-marketplace.netlify.app/demo/#numbered-headings)

disclaimer: this takes (heavy) inspiration from the [Mass.gov's process list](https://www.mass.gov/how-to/apply-for-unemployment-insurance-benefits) that @cmajel shared.

```html
<ol class="numbered-list">
  <li>
    <h3>Take the first step</h3>
  </li>
  <li>
    <h3>keep walking</h3>
  </li>
</ol>  
```

updating the guides manually would have been a slog but copilot made quick work of it in 1e09648. 👼 
Screenshot 2026-06-17 at 2 57 21 PM

### DIFF
--- a/src/demo.liquid
+++ b/src/demo.liquid
@@ -1,13 +1,13 @@
 ---
 layout: main
 stylesheet: guide.css
-
 title: Component demo
-description: >
-  A page to view individual layout components as they are completed
+description: A page to view individual layout components as they are completed
 dummyTags: ["GTFS-RT"]
 noindex: true
 show_call_to_action: true
+colorSchemes: ['yellow','purple','green','cyan-light','cyan-dark']
+
 ---
 <div class="container">
   <h1>DSDL component demo</h1>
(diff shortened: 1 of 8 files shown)


cmajel:
Tested this in VO + Chrome and VO + Safari and it looks good to me!

> taking this tack, the numbers are suppressed in the VoiceOver rotor, but overall i'd still call it an improvement.

This feels like a fine compromise to me but open to other opinions.

Scotchester:
> Tested this in VO + Chrome and VO + Safari and it looks good to me!
> 
> > taking this tack, the numbers are suppressed in the VoiceOver rotor, but overall i'd still call it an improvement.
> 
> This feels like a fine compromise to me but open to other opinions.

I agree on both counts! I had previously said I wasn't sure that not having the step number in the rotor's heading list was an acceptable trade-off, but on further consideration, if a user is stepping through the headings that way, I feel like the heading hierarchy provides _enough_ of that context (and it's less likely to be lost from one's head if they're not hearing all of the content between), and the strong improvement of how it reads in both Chrome and Safari is worth it.

jgravois:
awesome. thank you both!

src/demo.liquid
```diff
@@ -1,13 +1,13 @@
...
 stylesheet: guide.css
-
 title: Component demo
-description: >
-  A page to view individual layout components as they are completed
+description: A page to view individual layout components as they are completed
 dummyTags: ["GTFS-RT"]
 noindex: true
 show_call_to_action: true
+colorSchemes: ['yellow','purple','green','cyan-light','cyan-dark']
```
Scotchester, line 9: Hmm, should `show_call_to_action` also be camelCase?

src/styles/guide.css
```diff
@@ -26,8 +26,23 @@ ul > li:last-child {
   }
 }
 
-/* STEP LIST */
+/* STEP LIST and NUMBERED LIST */
```
Scotchester: I think it's fine for now, but if these ever graduate to DSDL components, I might like to revisit both the "step list" and "numbered list" names.

src/styles/guide.css
```diff
@@ -26,8 +26,23 @@ ul > li:last-child {
...
 
+.numbered-list,
+.step-list {
+  counter-reset: li;
+  list-style: none;
+  padding-inline-start: unset;
```
Scotchester: I'm not opposed to shifting toward logical properties, but maybe it'd be better to do it all at once in a given codebase? Seems a little odd for this to be the online one.

src/styles/guide.css
```diff
@@ -26,8 +26,23 @@ ul > li:last-child {
...
+}
+/* override bootstrap defaults for nested lists */
+.numbered-list ul {
+  list-style-type: disc;
+  margin-bottom: 1rem;
```
Scotchester:
```suggestion
  margin-bottom: var(--spacing-2);
```

src/styles/guide.css
```diff
@@ -26,8 +26,23 @@ ul > li:last-child {
...
+}
+.numbered-list h3 {
+  padding-bottom: var(--spacing-1);
+  padding-left: var(--spacing-10);
+}
```
Scotchester, line 49: I think we should keep these in `em`s in case the heading font size changes.

Sorta relatedly, the original implementation worked for headings of any level, but now it depends on them being `h3`s. Keepin' it YAGNI for now, I'd say it's another thing not to worry about it unless these graduate to DSDL components.

src/styles/guide.css
```diff
@@ -39,28 +54,29 @@ ul > li:last-child {
...
+.numbered-list > li,
+.step-list li {
   position: relative;
-  padding-top: 0.25em;
-  padding-bottom: 0.25em;
-  padding-left: 2.5em;
-  margin-top: var(--spacing-5);
+}
+.numbered-list > li {
+  margin-bottom: 1em;
```
Scotchester: This is unnecessary, given the `margin-top`.

```suggestion
```

src/styles/guide.css
```diff
@@ -39,28 +54,29 @@ ul > li:last-child {
...
+  margin-top: calc(var(--spacing-5) + 8px);
```
Scotchester: Is this to compensate for removing the top padding on the heading? I think I'd prefer to put that top padding back and remove this change and the `top: -8px` on the number, but I may be missing a good reason for doing it the way you have.

src/styles/guide.css
```diff
@@ -70,58 +86,55 @@ ul > li:last-child {
...
-  content: attr(data-number);
-  width: 1.75em;
+.numbered-list > li::before {
+  font-size: 2em;
   height: 1.75em;
   left: 0;
-  top: 0;
-  font-size: 1em;
   line-height: 1.75em;
+  top: -8px;
```
Scotchester: If we keep the negative top positioning technique, let's switch to `em`s in case of a user font size change:

```suggestion
  top: -0.25em;
```

src/styles/guide.css
```diff
@@ -70,58 +86,55 @@ ul > li:last-child {
...
   width: var(--spacing-4);
   height: var(--spacing-4);
   left: calc(-1 * var(--spacing-6));
   line-height: calc(2rem + (1rem / 16));
 }
-.numbered::before {
-  content: attr(data-number);
-  width: 1.75em;
+.numbered-list > li::before {
+  font-size: 2em;
```
Scotchester: Now that it's decoupled from the heading, we'll need a manual adjustment for the smaller `h3` font size on smaller screens:

A couple numbered headings on a phone-sized screen

Scotchester (changes requested):
Everything still reads well! Here's a smattering of observations on the code, some just notes for later work, but definitely want to fix the sizing on smaller screens.

Scotchester:
> updating the guides manually would have been a slog but copilot made quick work of it in [1e09648](https://github.com/cal-itp/mobility-marketplace/commit/1e096483c7ce2fa15eb470318eff32981a312e6a). 👼

Nice that it was smart enough to include the closing `</ol>` that you left out of your sample :)

src/demo.liquid
jgravois: i think it should! 1c421a2

src/styles/guide.css
jgravois: i agree wholeheartedly.

i couldn't think of an improvement that didn't involve renaming the step-list too, but i'm open to alternatives if anyone has any suggestions.

either in this PR or later.

src/styles/guide.css
jgravois: i have a question about this one that i think would be easier to ask you on the phone.

src/styles/guide.css
jgravois: great catch. fixed in 6617d7c

src/styles/guide.css
jgravois: for now i just left a code comment breadcrumb.

src/styles/guide.css
jgravois: fixed in 6617d7c

src/styles/guide.css
```diff
@@ -26,8 +26,24 @@ ul > li:last-child {
...
+/* STEP LIST and NUMBERED LIST (better naming would be welcomed :) ) */
 
+.numbered-list,
+.step-list {
+  counter-reset: li;
+  list-style: none;
+  padding-left: 0;
+}
+/* override bootstrap defaults for nested lists */
+.numbered-list > ul {
```
Scotchester: I realized after we discussed it that we'll actually need to go a bit further here. And let's include `ol`s while we're at it:

```suggestion
.numbered-list > li > ol,
.numbered-list > li > ul {
```

jgravois: great catch. when i verified this manually before i must have still been using `.numbered-list-item`.

0a01566 includes the spirit of your suggestion, but i don't think we want to override the `list-type-style` of ordered lists, so i broke that declaration out separately.

Scotchester: Ah yes, good point :)

Scotchester (approved):
LGTM!

PR #936, opened 2026-06-22
fix: ensure header link :focus style doesn't survive browser history navigation on mobile devices

### LINKED ISSUE #871: bug: hover style persists when navigating using 'Back'
via https://github.com/cal-itp/mobility-marketplace/pull/867#issuecomment-4173507222

> A minor thing on mobile. I noticed when I click a [card button](https://www.figma.com/design/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=10790-7897&m=dev) and then go back to home, the card button returns with a hover state. Is that something we can reset and have it just show the default style?

output

jgravois:
this change is 2/3 of the way toward resolving #871. 

AFAICT the only thing that will clear/reset an applied `:hover` style on desktop after history navigation is subsequent page interaction. however, omitting `:hover` styles altogether on devices that don't actually support it ensures that the conflated style doesn't stubbornly persist even _after_ the user starts to interact with the page.

before:
before

after:
after

components that have been re-treated:

1. homepage guide card links
1. header links
1. footer links
1. related guide links
1. view/download links

Screenshot 2026-06-23 at 9 33 06 AM
Screenshot 2026-06-23 at 9 33 18 AM
Screenshot 2026-06-23 at 9 33 48 AM
Screenshot 2026-06-23 at 9 34 23 AM

---

initially i just asked Gemini about this, but after i played around a bit with the fix it recommended i did a little more research of my own and found a really helpful article written by a human being: https://jacobpadilla.com/writing/hover-media-query

[`any-hover`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/any-hover) definitely was/is a new one for me.

~there are a few other locations in the CSS where we apply `:hover` styles, but I couldn't reproduce undesirable sticky behavior for any of them so I left all other declarations alone.~

### DIFF
--- a/src/styles/header.css
+++ b/src/styles/header.css
@@ -42,10 +42,17 @@
   outline-offset: var(--button-focus-outline-offset);
 }
 
-.global-nav a:hover,
-.header a.contact:hover {
-  border-color: var(--button-hover-border-color);
-  background-color: var(--button-hover-bg-color);
+/*
+this media query ensures that the user device has access to at least one input method that supports hover
+but the real reason for it is to side-step the propensity of some browsers to blend/persist
+:hover and :focus during history navigation and reapply the :hover style when "Back" is clicked
+*/
+@media (any-hover: hover) {
+  .global-nav a:hover,
+  .header a.contact:hover {
+    border-color: var(--button-hover-border-color);
+    background-color: var(--button-hover-bg-color);
+  }
 }
 
 .global-nav a:active,
--- a/src/styles/header.css
+++ b/src/styles/header.css
@@ -44,8 +44,8 @@
 
 /*
 this media query ensures that the user device has access to at least one input method that supports hover
-but the real reason for it is to side-step the propensity of some browsers to blend/persist
-:hover and :focus during history navigation and reapply the :hover style when "Back" is clicked
+the main reason for it is to ensure that mobile devices don't end up with a stuck :hover style on an element
+after the "Back" button is used for navigation
 */
 @media (any-hover: hover) {
   .global-nav a:hover,
--- a/src/styles/buttons.css
+++ b/src/styles/buttons.css
@@ -6,10 +6,16 @@
   font-size: var(--text-l);
 }
 
-.btn.btn-outline-primary:focus,
-.btn.btn-outline-primary:hover {
+.btn.btn-outline-primary:focus {
   border-color: var(--button-hover-border-color);
   background-color: var(--button-hover-bg-color);
+}
+
+@media (any-hover: hover) {
+  .btn.btn-outline-primary:hover {
+    border-color: var(--button-hover-border-color);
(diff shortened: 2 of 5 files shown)


Scotchester:
> there are a few other locations in the CSS where we apply `:hover` styles, but I couldn't reproduce undesirable sticky behavior for any of them so I left all other declarations alone.

I can reproduce it everywhere, it seems. For example, the big prev/next buttons at the bottoms of the guides, the view button on download cards, and footer links.

Also, this doesn't seem to fix it for me, which I think makes sense, since the styling on desktop isn't changing at all, and the article you linked (as well as others I found) seem to indicate that that technique is intended to address seeing undesirable hover styles on touch devices, but we're having the issue on non-touch devices.

I can't find any people commenting on the specific issue Christine identified, which is surprising. Seems like a lot of people would have noticed it. Although, maybe not, because in my testing the issue only presents if you use the browser back button to go back and keep your mouse off the content area. If people are using a back button on a mouse, or a keyboard shortcut, or a two finger gesture on a trackpad, their cursor is likely remaining in the content area, which is immediately resetting the hover styling.

I'm not sure this is solvable… 🤔

jgravois:
> Also, this doesn't seem to fix it for me

wow. 🤦‍♂️ 

it turns out w/ or w/o my bogus fix, the problem is not reproducible when running `npx @11ty/eleventy --serve` locally because the live reloading option sets the request header below. 😇 

```
Cache-Control: no-cache
```

that'll teach me to:
1. verify the misbehavior locally _prior_ to making changes
1. verify the fix on CI/CD immediately before opening a PR up for review

jgravois:
okay, i've doubled down on my original proposal even though its admittedly _not_ a silver bullet.

src/styles/header.css
```diff
@@ -42,10 +42,17 @@
...
 
-.global-nav a:hover,
-.header a.contact:hover {
-  border-color: var(--button-hover-border-color);
-  background-color: var(--button-hover-bg-color);
+/*
+this media query ensures that the user device has access to at least one input method that supports hover
+the main reason for it is to ensure that mobile devices don't end up with a stuck :hover style on an element
+after the "Back" button is used for navigation
+*/
```
Scotchester, line 49: I'm not sure if this comment is necessary anymore, given that we're doing the same thing across 5 files and certainly don't want to repeat it in every place it happens, but I don't feel strongly about it.

Scotchester (approved):
Tested with mobile Chrome and it works as expected!

src/styles/header.css
jgravois: yeah, i felt like it would be overkill to repeat the same comment everywhere we're doing this too.

i added it to the very first one and then chuckled when i saw that it ended up near the absolute bottom of the diff.

PR #937, opened 2026-06-22
Chore: add spellcheck to devcontainer, pre-commit, and CI checks

### LINKED ISSUE #870: Introduce spellcheck to devcontainer, pre-commit, and CI checks
We'd like to check for spelling errors in this repository, following the model established in the calitp.org repo (PRs cal-itp/calitp.org#548, cal-itp/calitp.org#563, and cal-itp/calitp.org#564).

lalver1:
Closes #870

This PR sets up a spell checker for this repo following the model established in the [calitp.org](https://github.com/cal-itp/calitp.org) repo. This means:

- Adding the [Code Spell Checker plugin](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker) to the devcontainer
- Adding the [cspell hook](https://github.com/streetsidesoftware/cspell-cli) to pre-commit
- Adding a [spell check GitHub Actions workflow](https://github.com/streetsidesoftware/cspell-action)

It also fixes a few typos/spelling errors that were found while adding these features.

## Reviewing

- For the VS Code plugin, edit an existing file with a spelling mistake, look at the `SPELL CHECKER` tab in the VS Code Panel, and check that the mistake was reported in the tab. Note that the plugin only works on the currently open file.
- To spell check the repo, run `npx cspell-cli --gitignore .` in your dev container terminal. This is roughly what the pre-commit hook and the `check-spelling` workflow will run, with the difference that pre-commit hook will check the commit message and only check on the staged files of the commit. Make any typo in a committed file and see the error appear in the output of the CLI. To see how the spell checker works in our CI, a [failing pre-commit workflow can be seen here](https://results.pre-commit.ci/run/github/288374589/1782225913.bLvjPqjbQHyeNN0sC3wmdg), and [a failing GH Actions workflow can be seen here](https://github.com/cal-itp/mobility-marketplace/actions/runs/27987980473/job/82833862503).


.cspell/dev-dictionary.txt
```diff
@@ -0,0 +1,59 @@
+# Words pertaining to development of this site
+
+aria
+booleancheckbox
+browserconfig
+btn
+calitp-
```
jgravois, line 7: i'm just being pedantic, but i would have expected to see "calitp-" in the `ignoreRegExpList` whitelist because we consider calitp itself a misspelling/mis-capitalization/mis-hyphenation and we want to treat this specific instance as a valid exception.

that said, i know we're doing the same thing right now in [calitp.org](https://github.com/cal-itp/calitp.org/blob/25386537d146baa5257642e356af1f7070da6e60/.cspell/dev-dictionary.txt#L3) and fundamentally i don't think there's really much of a difference between the two.

maybe a better way to think of it is that `ignoreRegExpList` should be reserved for actual regexes and plain strings should just be added to the dictionary??

haha. thanks for coming to my TED talk.

lalver1: No, good point @jgravois. Working on this PR I realized that most of the words in the custom dictionaries are easy to categorize where they should be listed so that they are ignored, but for some, like `calitp-` they could arguably be listed in different places. For this one in particular, I think it mostly shows up in `css` files, so that's probably why it's listed in the `dev-dictionary` (similarly for the calitp.org repo, as you noted).

Yep, I ended up thinking about `ignoreRegExpList` as strictly only containing regexes, which usually are used to keep the dictionaries from getting too bloated with variations of a word, or when the word we are interested in ignoring ends up showing up across different file types so we want to make sure we catch it wherever it appears.

jgravois: > I ended up thinking about ignoreRegExpList as strictly only containing regexes

i think that's an exceedingly reasonable way to look at it. by that logic it makes sense (to me) to move all the plain strings below that can actually be found in this repo out of `ignoreRegExpList` and directly into the dictionaries, but i definitely wouldn't call that a blocker.

```
    "/Cal-ITP/",
    "/.ITP/",
    "/github.com\\/cal-itp/",
    "/cal-itp\\/calitp.org/",
    "/\/usr\/src\/calitp/",
    "logo-filenames",
    "google-analytics-id"
```

Scotchester: I can't remember the exact specifics off the top of my head, but I think there are two potential problems with that approach (for _some_ of the words in the `ignoreRegExpList`):

1. I think the dictionaries may not be case-sensitive? Which matters for some of the flavors of Cal-ITP.
2. I think they may not support certain punctuation (slashes, in particular, look like they're in there because they're problematic, maybe hyphens, too).

If you want to play around and try to get to something that seems better, by all means. I'd recommend putting a temporary list of things you do/don't expect to get flagged in the README or something and just check it live as you make updates.

Scotchester: Not sure it completely addresses your questions/proposal, but some of the logic is described in my original calitp.org spellcheck PR: https://github.com/cal-itp/calitp.org/pull/548

jgravois: helpful context @Scotchester. i'd sign off on this as-is even if `ignoreRegExpList` isn't serving a special purpose. 🤗

jgravois:
> To spell check the repo, `run npx cspell-cli --gitignore`

since this repo has a package.json file we _could_ canonize `cspell-cli` as a devDependency and create an npm script to spellcheck the entire repo.

i don't think we should bother with any of this today but the whole topic has me thinking about the 🤸 we're doing to keep the version of prettier VS Code is using in sync with the version pre-commit loads (example: #927). it sure would be cool if we could consolidate.

```yaml
repos:
  - repo: local
    hooks:
      - id: spellcheck
        name: Spellcheck (via npm)
        entry: npm run spellcheck
        language: node
```

jgravois (approved):
nice work @lalver1!

PR #938, opened 2026-06-23
Add unique Open Graph images for the guides

### LINKED ISSUE #935: Implement distinct image previews for individual guides
in #796 @cmajel curated new social media images for each of our guides

## Acceptance Criteria

- [ ] when a link to any of our guides (`/data-plans/`, `/route-scheduling/`, ...) is shared via slack, facebook, RCS, etc. they include a unique thumbnail image.

```html
<meta property="og:image" content="https://www.camobilitymarketplace.org/uploads/og-image.png">
```

since right now the relevant meta tags ☝️ are a FQDN, we'll have to either:

1. open a single PR and wait until after it is merged in order to test the new behavior

_or_

1. open one PR to check the new images into version control and open a second PR to update the tags.

personally, i'm fine with either option.

Scotchester:
Closes #935 

---

This also sets up a `siteDomain` global data item that should theoretically dynamically show the Netlify deployment domain regardless of environment. We will have to test this on the actual preview deployment for this PR.


Scotchester:
It works! 🎉

jgravois (commented):
rockin! 🎸 

it feels 5% cleaner to me to manage this variable alongside the others instead of leaning on `addGlobalData`, but i'm okay with what you're proposing too.

```js
// site.js
export default {
  // ...
  domain: process.env.DEPLOY_PRIME_URL || "http://localhost:8080";
}
```

either way, i think it'd behoove us to avoid maintaining both `site.url` _and_ `siteDomain`.

Scotchester:
> rockin! 🎸
> 
> it feels 5% cleaner to me to manage this variable alongside the others instead of leaning on `addGlobalData`, but i'm okay with what you're proposing too.
> 
> ```js
> // site.js
> export default {
>   // ...
>   domain: process.env.DEPLOY_PRIME_URL || "http://localhost:8080";
> }
> ```
> 
> either way, i think it'd behoove us to avoid maintaining both `site.url` _and_ `siteDomain`.

Great call. I followed the advice I found without thinking critically enough about it.

Done in 42e2456.

src/_data/site.js
```diff
@@ -14,5 +14,5 @@ export default {
   theme_color: "#ffffff",
   theme_color_background: "#ffffff",
   title: "California Mobility Marketplace",
-  url: "https://www.camobilitymarketplace.org",
+  domain: process.env.DEPLOY_PRIME_URL || "http://localhost:8080",
```
angela-tran: Do you think we should add in ` || site.url` before the fallback to localhost, just in case `process.env.DEPLOY_PRIME_URL` somehow ends up empty?

angela-tran (dismissed):
This looks great @Scotchester! I like how this all looks with `site.domain`.

Just one non-blocking suggestion for you to take or leave

src/_data/site.js
jgravois: i'm fine with this as-is too, but i could also get behind differentiating the fallback using the build mode.
```js
const domain = process.env.ELEVENTY_RUN_MODE === "build"
  ? process.env.DEPLOY_PRIME_URL || "https://www.camobilitymarketplace.org"
  : "http://localhost:8080"
```

Scotchester: Yeah, that is probably smart, Angela, though `site.url` no longer exists. John's approach seems like a good one to me. Making that change now!

Scotchester:
Local env and deploy preview still look good to me. Need one more approval, please!

cmajel:
🎉 I don't have approval powers but the previews look great!

jgravois:
> I don't have approval powers but the previews look great!

@cmajel i'm surprised to hear that because you're a member of both [@web-admin](https://github.com/orgs/cal-itp/teams/web-admin) and [@web-review](https://github.com/orgs/cal-itp/teams/web-review).

what do you see in the "Submit review" dialog?

cmajel (approved):

cmajel:
@jgravois pointed me in the right direction (I could have sworn I used to be able to submit reviews from the conversation tab)

PR #942, opened 2026-06-24
chore: cache node install in devcontainer

### LINKED ISSUE #941: chore: ensure node installation survives a devcontainer rebuild
the way things are set up right now Node.js isn't installed until the postAttach.sh script is run.

it would be more performant to ensure this happens within the Dockerfile itself so that the step is cached for subsequent builds.

prior art: https://github.com/cal-itp/calitp.org/pull/663

jgravois:
resolves #941 

i didn't notice it when i was working on https://github.com/cal-itp/calitp.org/pull/663, but one side-effect of mounting 'node_modules' as a volume is that we can't just nuke it from within the devcontainer in the event that we genuinely want to start from scratch.

Screenshot 2026-06-24 at 3 55 58 PM

i doubt it'll come up often in practice, but a quick chat with Gemini clued me into the fact that it is possible to use `find` to imitate the behavior of `rm -rf node_modules` and leave the parent directory intact.

```bash
find ./node_modules -mindepth 1 -delete
```

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -3,22 +3,25 @@ FROM python:3.12
 ENV USER=mobimart
 ENV HOME=/home/$USER
 
+ENV WORK_DIR=$HOME/site \
+    MISE_DIR=$HOME/mise
+
 RUN useradd --create-home --shell /bin/bash $USER && \
+    mkdir $WORK_DIR && \
     chown -R $USER $HOME
 
-WORKDIR $HOME/site
+WORKDIR $WORK_DIR
 
 # Install mise (to manage Node.js) https://mise.jdx.dev/mise-cookbook/docker.html#docker-image-with-mise
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
-ENV MISE_DATA_DIR="$HOME/mise"
-ENV MISE_CONFIG_DIR="$HOME/mise"
-ENV MISE_CACHE_DIR="$HOME/mise/cache"
+ENV MISE_DATA_DIR=$MISE_DIR
+ENV MISE_CONFIG_DIR=$MISE_DIR
+ENV MISE_CACHE_DIR="$MISE_DIR/cache"
 ENV MISE_INSTALL_PATH="/usr/local/bin/mise"
-ENV PATH="$HOME/mise/shims:$PATH"
+ENV MISE_TRUSTED_CONFIG_PATHS=$WORK_DIR
+ENV PATH="$MISE_DIR/shims:$PATH"
 
 RUN curl https://mise.run | sh
-
-RUN apt-get update
 
 # Install Python dependencies
 COPY .devcontainer/requirements.txt .devcontainer/requirements.txt
@@ -28,3 +31,7 @@ RUN pip install --no-cache-dir -r .devcontainer/requirements.txt
 USER $USER
 
 RUN git config --global --add safe.directory $HOME
+
+# Install Node.js and npm dependencies
+COPY mise.toml .node-version package.json package-lock.json ./
+RUN mise exec -- npm install -g npm@latest && npm clean-install
--- a/.devcontainer/compose.yml
+++ b/.devcontainer/compose.yml
@@ -6,3 +6,4 @@ services:
     command: sleep infinity
     volumes:
       - ..:/home/mobimart/site
+      - /home/mobimart/site/node_modules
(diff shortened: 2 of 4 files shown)


.devcontainer/Dockerfile
```diff
@@ -3,23 +3,26 @@ FROM python:3.12
 ENV USER=mobimart
 ENV HOME=/home/$USER
 
+ENV WORK_DIR=$HOME/site \
+    MISE_DIR=$HOME/mise
```
Scotchester, line 7: I was wondering about the inconsistency of whether we group `ENV` declarations or not, and then I realized it's probably because using one env var to define another requires that it's already been set by a previous statement. I believe this line in [the Dockerfile reference docs](https://docs.docker.com/reference/dockerfile/#env) confirms that:

> A stage inherits any environment variables that were set using ENV by its parent stage or any ancestor.

However, reading that section of the docs also leads me to the observation that these vars are only used in the Dockerfile, and so I think they could be `ARG`s that don't persist as env vars in the final image.

Scotchester (commented):
Working for me! One non-blocking comment.

.devcontainer/Dockerfile
jgravois: > it's probably because using one env var to define another requires that it's already been set by a previous statement.

that's my assumption too.

> I think they could be ARGs that don't persist as env vars in the final image.

interesting idea. i'll let sleeping dogs lie here since none of the values contain sensitive, but i think its worth keeping this tidbit in mind going forward for sure.

jgravois:
> Working for me! One non-blocking comment.

auto-merge enabled. 😇

Scotchester:
Ah sorry, I had meant to approve, but @lalver1 mentioned wanting to review this also, so we can wait for him.

.devcontainer/Dockerfile
thekaveman: >> it's probably because using one env var to define another requires that it's already been set by a previous statement.

> that's my assumption too.

Indeed this is the case. If you have the `docker.docker` VS Code extension (`Docker DX`) installed, you will see this warning if attempting to use an ENV before it is fully defined:

image

Scotchester: Should we install that extension in all of our dev containers?

Scotchester: Or Microsoft's [Container Tools](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-containers) extension, which looks like it provides some of its own Dockerfile helpers?

jgravois: in my case it would have been more helpful to have it installed _outside_ the devcontainer.

Scotchester: A quick check of running Container Tools without Docker DX indicates that it isn't sophisticated enough to catch that `Usage of undefined variable` error, so maybe both are warranted.

thekaveman: Right, `Docker DX` is specifically for highlighting best practices etc.

> Should we install that extension in all of our dev containers?

I don't know if it makes sense _within_ the devcontainer? Since you can't build the Docker image in there. I typically do all my Docker editing/building _outside_ a devcontainer, where this extension is more useful.

lalver1 (approved):
This works for me too! 👍 

Just a few notes, mostly stuff that's fairly new to me and that's different enough from how `pip` does things.

- I was wondering if it's possible for `npm` to create `node_modules` somewhere else besides the project's workspace to avoid this quirk of mounting the host workspace in the dev container and it overwriting the installed packages. But it looks like it's not recommended since many tools assume `node_modules` to be in its default location. We don't see this quirk in Python projects because `pip` installs packages in places like `/usr/.../site-packages` outside of the project's workspace. I guess we'd be in the same situation for Python projects if we used the default setup of a virtual environment created in the project workspace, but there's no need for virtual environments in our dev containers.
- I'll keep in mind the command about how to completely delete the `node_modules` folder @jgravois. But like you said, that is not something we do often. Usually, like if we need to review a PR, we'll just update a dependency in `package.json` (and have the lockfile updated too) and test out the changes. To be extra sure, we can also do a "Rebuild and reopen in container" to ensure the updated version of a dependency is used. I tested this scenario out and when I ran `npm list` I saw the versions that I was expecting 👍 .

jgravois:
> Usually, like if we need to review a PR, we'll just update a dependency in package.json (and have the lockfile updated too) and test out the changes.

that's exactly right @lalver1.

to take a new dependency for a test drive, you'd do 'npm install --save-dev foo` from within the devcontainer.

to test out a dependency change in someone else's PR you could either rebuild the container (i don't _think_ its necessary to clear the cache) or run `npm clean-install` from within the devcontainer.

PR #958, opened 2026-07-29
Add broken link checker

### LINKED ISSUE #962: [spike] Link scanner tool
As part of transitioning Cal-ITP websites to Caltrans, we will need documentation and tools to support keeping content on both websites relevant and working. 

This ticket covers research into whether there is utility in employing a link scanner tool (perhaps a feature in 11ty or a plugin to check that links still resolve at least to a 200 / good response vs. a 404 etc. kind of response.

Scotchester:
Part of cal-itp/mobility-marketplace#962

---

- [x] Check PRs that update the site
- [x] Run on a schedule to catch external links that change out from under us


jgravois (commented):
looks promising!

i'll hold off on reviewing this one until i see a completed workflow run.

Scotchester:
Thought I had left this in draft, but apparently not. It's now ready for review!

For the PR check, I elected to just have it run like any other PR check and report success or failure (i.e., it doesn't post a comment as I previously suggested doing). You can see it doing this on this very PR.

For the weekly run, the approach I took was to have it open an issue with a copy of the report if there are any failures. If it finds that there is already an issue with the `linkchecker` label on it, it will add a comment to that issue with the latest results, rather than creating a new issue, so will only be one `linkchecker` issue open at a time. The idea was to prevent a proliferation of issues reporting the same problems if we can't remediate them all in the space of a week.

You can see an example of how it looks in the test repo: https://github.com/compilerla/actions-testing/issues/166

Feedback welcome!

.github/workflows/check-links-prs.yml
```diff
@@ -0,0 +1,47 @@
+name: Check links
+
+on:
+  pull_request:
+    paths:
+      - "eleventy.config.js"
+      - "linkcheckerrc"
+      - "src/**"
```
jgravois, line 8: with this many path filters, i'd be inclined to just yank `paths` altogether.

.github/workflows/check-links-prs.yml
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  linkchecker:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5 # prevent background server step from running forever if something else goes wrong
```
jgravois, line 13: this reminds me that it would be cool to have our new `terraform` gh actions self-destruct _and_ ensure their associated locks are removed....

.github/workflows/check-links-weekly.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Strip all but body contents
+        id: extract-report-body
+        if: ${{ failure() }}
+        run: python3 -c 'from bs4 import BeautifulSoup; print(BeautifulSoup(open("results.html").read(), "html.parser").body.decode_contents())' > report_body.html
+
+      - name: Check for an open linkchecker issue
+        id: open-issue-check
+        if: ${{ failure() }}
+        run: echo "OPEN_ISSUE_ID=`gh issue list --label linkchecker --json number --jq '.[0].number'`" >> "$GITHUB_OUTPUT"
```
jgravois, line 36: i dig that you're just using `gh` for all this.

.github/workflows/check-links-weekly.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Add content to existing issue
+        if: ${{ failure() && steps.open-issue-check.outputs.OPEN_ISSUE_ID != '' }}
+        run: |
+          echo -e "## Updated broken link report\n\nFrom weekly run: https://github.com/cal-itp/mobility-marketplace/actions/runs/${{ github.run_id }}\n\n<details>\n" > comment.md
+          cat report_body.html >> comment.md
+          echo -e "\n</details>" >> comment.md
+          gh issue comment ${{ steps.open-issue-check.outputs.OPEN_ISSUE_ID }} \
+            --body-file comment.md \
+            --repo $GITHUB_REPOSITORY
```
jgravois, line 46: YMMV, but in [`peter-evans/create-or-update-comment`](https://github.com/peter-evans/create-or-update-comment#action-inputs), the default edit-mode mode _appends_ files to anything existing it finds...

linkcheckerrc
```diff
@@ -0,0 +1,6 @@
+[filtering]
+ignore=
+  /cdn-cgi/l/email-protection
```
jgravois, line 3: what are these links pray tell?

jgravois (commented):
i have some feedback that is orthogonal to something i said inline.

personally, i think the cron workflow is substantially higher value here than the action that runs on `pull_request`. if we aren't catching broken links during manual QA, we have bigger problems.

linkcheckerrc
Scotchester: Turns out that we have Cloudflare set up to mask email addresses, so the `href` that `linkchecker` sees for a `mailto` link looks like this:

```html
<a href="/cdn-cgi/l/email-protection#79111c151516391a1815100d0957160b1e">
  <span class="__cf_email__" data-cfemail="670f020b0b082704060b0e131749081500">[email&#160;protected]</span>
</a>
```

And those result in 404s if it just tries to follow them directly.

jgravois: _very_ cool!

thekaveman: I literally had no idea... :thinking:

.github/workflows/check-links-weekly.yml
Scotchester: Thanks for suggesting this. I have a few reasons for preferring the current approach, though:

- We have expressed a desire to reduce our allowance on third-party actions.
- It's not even clear to me from the docs/issues in that repo if that action can update the original issue description, or just comments. Is the description considered a comment that can be acted on in the same way? Would require some deeper research and/or testing.
- Most importantly (to me), editing the issue description/comment will not create a new notification for those watching the issue, which will make it harder to notice new results that occur after someone has already started working on the prior results (or make it easy to forget how long reported issues have going unaddressed).

.github/workflows/check-links-weekly.yml
Scotchester: It's a pretty powerful addition to the built-in GHA toolset!

.github/workflows/check-links-weekly.yml
jgravois: sorry. i didn't intend to suggest that you switch to the github action i linked to.

i was only saying that the underlying `gh` behavior might be the same. 😇

.github/workflows/check-links-prs.yml
Scotchester: I really prefer not to run unnecessary checks when we can avoid them without too much effort. I don't think we actually have to pay for Actions minutes since our repos are open source, but it's still wasteful and adds noise on PRs where they don't apply.

.github/workflows/check-links-weekly.yml
Scotchester: Ah!

I wouldn't expect that to be the case since I'm just telling it to make a new comment and not referencing an exiting comment, but I'll run the workflow in the test repo manually again and confirm what happens.

Scotchester: Confirmed: https://github.com/compilerla/actions-testing/issues/166#issuecomment-5198195276

.github/workflows/check-links-prs.yml
jgravois: that's why my orthogonal suggestion was to get rid of the workflow altogether! 😆

Scotchester: That's a solid point! I'm not opposed to not doing it on PRs of everyone agrees a weekly check is sufficient.

thekaveman: Yep I'm good with the weekly

thekaveman (changes requested):
Just to make it explicit, we are awaiting removal of the PR workflow and aligned on the weekly automated workflow only 👍

Scotchester:
PR workflow removed, ready for final review!

jgravois (approved):
📚

.github/workflows/check-links-weekly.yml
```diff
@@ -0,0 +1,56 @@
+name: Weekly Broken Link Check
+
+on:
+  schedule:
+    - cron: "0 9 * * MON"
```
thekaveman: You can add a `timezone` param here to make this 9am on Mondays on the West Coast (or in other words, make it more explicit what time this should actually run):

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#schedule

Scotchester: I did mean for it to be middle of the night for all of America, but yeah, I'll make that more explicit :)

jgravois (approved):
🗺️

PR #961, opened 2026-08-05
fix: Update logic in site.domain for proper results in prod

### LINKED ISSUE #960: Sitemap and other URLs use netlify.app domain instead of proper one
I discovered when testing the broken link checker that the sitemap at https://www.camobilitymarketplace.org/sitemap.xml is using `main--cal-itp-mobility-marketplace.netlify.app` as the hostname in all of its URLs. It's also being used in our canonical link and Open Graph tags.

At a glance, this was caused by some mistake I made in #938. I said:

> This also sets up a siteDomain global data item that should theoretically dynamically show the Netlify deployment domain regardless of environment. We will have to test this on the actual preview deployment for this PR.

And it did indeed work on the preview deployments, but I failed to ensure the desired behavior still happened in production.

I can see where the logic went wrong and will implement a fix ASAP.

Scotchester:
Fixes #960 

---

This definitely works for localhost and ~~we'll see momentarily if it still works for~~ preview deploys.

@jgravois Could I ask you to double-check the ternary syntax here? This is not exactly how I would prefer to format it, but it's what Prettier wants.

### DIFF
--- a/src/_data/site.js
+++ b/src/_data/site.js
@@ -15,7 +15,9 @@ export default {
   theme_color_background: "#ffffff",
   title: "California Mobility Marketplace",
   domain:
-    process.env.ELEVENTY_RUN_MODE === "build"
-      ? process.env.DEPLOY_PRIME_URL || "https://www.camobilitymarketplace.org"
-      : "http://localhost:8080",
+    process.env.ELEVENTY_RUN_MODE === "build" // if 11ty is building
+      ? process.env.BRANCH === "main" // and the branch is main
+        ? "https://www.camobilitymarketplace.org" // use the production domain
+        : process.env.DEPLOY_PRIME_URL // otherwise use Netlify's deploy domain
+      : "http://localhost:8080", // but if 11ty is serving, assume localhost
 };


thekaveman:
@Scotchester 

> This is not exactly how I would prefer to format it

This is just a `.js` file right? An expression could be written outside of this `export default` block if you wanted a different format. Not worried either way.

jgravois (approved):
+1 to @thekaveman's suggestion about teasing out a few declarations, but this nested ternary is plenty legible as-is to a kool-aid drinker like me (especially with all your great code comments).

```js
const { BRANCH, DEPLOY_PRIME_URL, ELEVENTY_RUN_MODE } = process.env

const IS_BUILD = ELEVENTY_RUN_MODE === "build"
const IS_MAIN = BRANCH === "main"

const LOCAL = "http://localhost:8080"
const CANONICAL = "https://www.camobilitymarketplace.org"

// ...
  domain: IS_BUILD ? IS_MAIN ? CANONICAL : DEPLOY_PRIME_URL : LOCAL
```

Scotchester:
Great point about it being just a JS file, Kegan, and thanks for the reassurance, John.

In that case, I'll leave it as is. I think I actually prefer what I have now to all those constants, which sorta make my eyes glaze over in the ensuing one-line `domain` ternary :)

